PR #5, opened 2017-04-13
eliminate reference to nonexistent `object`

bgoodri:
The Gaussian example in README currently does not run because `proj_coef_helper` tries to access `object` which does not exist. With this PR, the example works again.


jpiironen:
Thanks for the fix. Among many of the changes to the software during the past couple of weeks, we ended up removing this functionality (at least temporarily), perhaps including a better designed version of it later, and thus the PR is deprecated. So I will close this for now but thanks anyway!

PR #13, opened 2018-04-01
Compatibility with loo v2.0.0

jgabry:
@jpiironen I think this is what’s needed to fix things to be compatible with the new loo package.


avehtari:
Hi @jgabry, we discussed loo 2 with @jpiironen last week and we'll check this tomorrow (there's a 4 day Easter holiday in Finland)

jgabry:
Ok great

jpiironen:
This seems correct. We thought it would be a good idea to wait until loo 2.0 is in CRAN and then merge this to avoid possible problems if someone installs projpred from Github. Can you inform me when the new loo is in CRAN? Thanks.

jgabry:
Ok cool. I’ll definitely let you know when loo is updated. I just finished
editing the vignettes, which I think was the last thing to do before
submitting. So hopefully I’ll submit tomorrow.

PR #18, opened 2019-05-16
Remove rngtools

### LINKED ISSUE #17: rngtools package now depends on R >= 3.6.0
Unfortunately, as of the latest release of **rngtools** (which is listed in Imports for **projpred**) it now depends on R >= 3.6.0 (see https://cran.r-project.org/web/packages/rngtools/index.html), but **projpred** is listed as depending only on R >= 3.1.2. 

https://github.com/stan-dev/projpred/blob/b2a8f92f22fc246d06de663f7a8393b9d8e86a1b/DESCRIPTION#L17-L18

Because **projpred** Imports **rngtools**

https://github.com/stan-dev/projpred/blob/b2a8f92f22fc246d06de663f7a8393b9d8e86a1b/DESCRIPTION#L24

this means we should either require R >= 3.6.0 for **projpred** too or drop the dependency on **rngtools**. If possible we should do the latter because there's really no reason for **projpred** itself to require the latest R version. 

Is using **rngtools** definitely necessary? I think we may be able to do everything without it.

jgabry:
Closes #17 

This PR attempts to remove uses of the **rngtools** package. All of the tests pass for me when I run this.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -20,8 +20,7 @@ Imports:
     loo (>= 2.0.0), 
     ggplot2, 
     Rcpp, 
-    utils,
-    rngtools (>= 1.2.4)
+    utils
 LinkingTo: Rcpp, RcppArmadillo
 License: GPL-3
 Encoding: UTF-8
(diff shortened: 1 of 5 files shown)


jpiironen:
Looks good to me. All the tests run fine for me too. I merged now, let's fix later if we find something.

PR #20, opened 2019-06-25
Check the length of the penalty vector (fixes #19).

### LINKED ISSUE #19: check the length of the penalty vector
Currently the `penalty` argument of `varsel()` is not checked to be of the expected length, which raises a confusing error and some warnings.

```{r}
 > vs <- varsel(fit, method='l1', penalty=c(1, 0))

Error in seq.default(log(lambda_min), log(lambda_max), len = nlam) :
  'from' must be a finite number
In addition: Warning messages:
1: In max(lambda_max_cand[is.finite(lambda_max_cand)]) :
  no non-missing arguments to max; returning -Inf
2: In log(lambda_min) : NaNs produced
3: In log(lambda_max) : NaNs produced
```

mcol:
This is a minimal approach to fix #19.


jpiironen:
Thanks for the PR. Looks good to me.

PR #22, opened 2019-06-28
Store the variable selection method in the varsel/cv_varsel objects

mcol:
This makes it straightforward to see what selection  method was used, especially helpful when similar objects are produced using different methods.


jpiironen:
Seems OK to me.

PR #23, opened 2019-06-28
Fix the error message if an unknown cv_method is specified

mcol:
I stumbled into this by chance:
```
> cvs <- cv_varsel(fit, cv_method="k-fold")
Error in cv_varsel(fit, cv_method = "k-fold") :
  Unknown cross-validation method: L1.
```


jpiironen:
Thanks for the fix. Merged.

PR #24, opened 2019-07-03
Check the method argument in parseargs_varsel() instead of select()

mcol:
This avoids some unnecessary work (such as the calls to `.get_traindata()` and `.get_refdist()`) if the method passed is neither "forward" nor "L1", so the call to `varsel()` or `cv_varsel()` will stop immediately rather than after a few seconds (or more for a large model).


jpiironen:
Seems ok.

PR #26, opened 2019-07-09
standard errors from varsel_stats() for stat='rmse' not reproducible

mcol:
If varsel_stats() is called with options stat='rmse' and delta=FALSE, the computation of the standard errors via bootstrap would not be exactly reproducible because the random seed is not set.


jpiironen:
Thanks for the fix.

Issue #25, opened 2019-07-09
Add AUC as option to varsel_stats()


mcol:
Currently only the "acc"/"pctcorr" option is specific for a binary outcome, and it's not great. Despite its limitations, the AUC is defintely better. I'm preparing a PR for this, which will introduce the "auc" stat.

avehtari:
Thanks for working on this. I would suggest adding Brier score, too (as it would be better than AUC)

mcol:
But the Brier score is effectively already computed with option "mse", so unless I'm missing something, it should be just a matter of adding a new option to those accepted by varsel_stats/varsel_plot and update the documentation. I could do that in a separate PR if that's what you mean.

jpiironen:
This is now implemented by #27.

PR #27, opened 2019-07-09
add AUC as option to varsel_stats() 

mcol:
This addresses #25. The AUC results are consistent to those produced by the WeightedROC package, which helped me to sort out the issue of ties in predicted probabilities. It's worth double-checking the computation of bootstrapped standard errors, especially in the `!is.null(mu.bs)` case (when `varsel_stats()` is called with `delta=TRUE`).

### DIFF
--- a/R/methods.R
+++ b/R/methods.R
@@ -221,6 +221,7 @@ proj_predict <- function(object, xnew, offsetnew = NULL, weightsnew = NULL,
 #'  \item{mse:} {Mean squared error (gaussian family only)}
 #'  \item{rmse:} {Root mean squared error (gaussian family only)}
 #'  \item{acc/pctcorr:} {Classification accuracy (binomial family only)}
+#'  \item{auc:} {Area under the ROC curve (binomial family only)}
 #' }
 #' Default is elpd.
 #' @param type One or more items from 'mean', 'se', 'lower' and 'upper' indicating which of these to
--- a/R/misc.R
+++ b/R/misc.R
@@ -54,15 +54,17 @@ bootstrap <- function(x, fun=mean, b=1000, oobfun=NULL, seed=NULL, ...) {
   rng_state_old <- .Random.seed
   on.exit(assign(".Random.seed", rng_state_old, envir = .GlobalEnv))
   set.seed(seed)
-  
+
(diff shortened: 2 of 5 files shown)


jpiironen:
Thanks, this is a good idea and your contribution is very much appreciated. I did not yet delve into the details of how this is computed, but immediately ran into an error when tried using this.
```
n <- 10
x <- matrix(rnorm(10*5), nrow=10)
y <- rbinom(10,1,0.5)
ref <- init_refmodel(x,y,family = binomial())
vsel <- cv_varsel(ref)
varsel_plot(vsel, stats='auc')
```
Here ```varsel_plot``` throws an error ``` Error in x[bsind, ] : subscript out of bounds  ```. You get the same if you call ```varsel_stats```.

In addition if you call ```suggest_size(vsel, stat = 'auc')``` you get ```Error in .is_util(stat) : Internal error: non-recognized statistic 'auc' ```. This latter one should be easy to fix, though, as you need only to add 'auc' in the list of ```recognized_stats``` in function ```.is_util ``` (This is a bit ugly to have such a definition inside a function, but we were not sure what would be the best choice to solve this, global variables are ugly too..).

Would you mind adding tests that would have caught these? I now just realized that we did not seem to have a test for testing ```suggest_size```, would be great if you could write a test that loops through all the possible stats and call ```suggest_size```, so then at least some sort of test for all stats would exist. Thanks!

mcol:
Thanks for your testing, I was mainly using rstanarm, and assumed the test were sufficients. I'll add a few more to validate inputs and for `suggest_size` in a separate PR.

The problem with the subscript out of bounds error has to do with how I tried to work around the fact that `bootstrap` only applies subsetting to the first argument, but the computation of the AUC requires three, so the correct approach will have to be more involved.

mcol:
The tests I wrote in #30 would have caught the above errors. When that's merged, I'll update this.

mcol:
It turns out that the problem was not with `bootstrap`, althought the error was raised there, but in how I was removing rows with missing values in the `auc.data` matrix. The approach I'm using now follows what's done in the 'rmse' case: I'm not sure I really understand what's going on, but the tests now pass.

jpiironen:
Thanks. Seems good to me, so merged.

PR #30, opened 2019-07-11
Improve tests for variable selection related functions

mcol:
This cleans up a bit `test-varsel.R`, adds a `.validate_vsel_object_stats` to consolidate the checks that the object and the stats argument is valid, and adds a few sanity checks for `suggest_size`.


jpiironen:
Thanks, this seems good.

Issue #42, opened 2020-02-27
Fix approach to search_L1


AlejandroCatalina:
This method involves a lot of old code that still does things very manually, so we have to find a better way to abstract the method. For the moment we'll at least extract the response safely by using `model.matrix`.

AlejandroCatalina:
Partially addressed in latest develop. I would still like to find a cleaner way
(closed for the moment).

PR #43, opened 2020-02-27
properly deal with response in predict.subfit

AlejandroCatalina:
This PR addresses #42 . Now we retrieve the data matrix in a safe way by using `model.matrix`.

I'm not sure if this should stay like this for the time being. We can leave this open until I investigate how `glmnet` exactly differs from the current approach.

ping @jpiironen if you remember why the current method is different and how so?

### DIFF
--- a/R/formula.R
+++ b/R/formula.R
@@ -293,11 +293,11 @@ subset_formula_and_data <- function(formula, terms_, data, y=NULL, split_formula
       formula <- lapply(response_cols, function(response)
         update(formula, paste0(response, " ~ .")))
     }
+    data <- data.frame(y, data)
   } else {
     formula <- update(formula, paste(response_cols, "~ ."))
   }
 
-  data <- data.frame(y, data)
   colnames(data)[seq_len(response_ncol)] <- response_cols
   return(nlist(formula, data))
 }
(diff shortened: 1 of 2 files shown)


jpiironen:
I don't remember exactly if there is a difference between projpred's `glm_elnet` function and `glmnet`, but I remember that at least at some point the lambda values did not match exactly (lambda = L1-regularization parameter). This "problem" might have gone away after some changes in the code but I don't remember anymore.

One reason that might/will cause difficulties is that if the features x are normalized differently internally, because the normalization has a big effect on the solution with a given nominal lambda value. `glmnet` explicitly discourages user to provide the lambda values, but instead encourages to provide only the number of lambda values (letting the function decide the appropriate lambda grid, which is a kinda subtle issue).

If you want to delve into this, I would recommend setting up some simple toy example (e.g. something like used in the tests), then run the L1-path using `glmnet`, then fetch the lambda grid from that object, and see `glm_elnet` gives the same result. I can also recommend reading through the original paper https://www.jstatsoft.org/article/view/v033i01, it's pretty well written in my opinion, and you will probably learn a lot by reading it (at least I did). My c++ implementation also follows that algorithm with only some minor tweaks (such as termination criterion etc.).

AlejandroCatalina:
I see. I have worked with glmnet before so I'm familiar with it. I'll take a look at this, although it's not the highest priority right now. Thanks!

paul-buerkner:
Merged after discussion with @AlejandroCatalina

PR #48, opened 2020-03-16
Refactor extend family

AlejandroCatalina:
Change `dbinom(y * weights, weights)` to `dbinom(y, weights)`.

This involves some light changes in `get_refmodel` and `.get_standard_y`.
At some point we should move `get_refmodel.stanreg` to `rstanarm`, and update `get_refmodel.brmsfit` in `brms` @paul-buerkner (now you don't need to return the target divided by the weights).


paul-buerkner:
You can merge it if you want.

PR #51, opened 2020-03-25
Improve non-Gaussian projections

AlejandroCatalina:
Right now we solve pure ML estimates for all approximations, even for
non-Gaussian ones, what may end up in unstable or degenerate estimates for some
models. By using ridge regression we can control for this, but we need to add
MASS dependency to use lm.ridge (we want to use this to have a similar
interface to lme4).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,4 +28,4 @@ after_success:
 script: 
   - |
     R CMD build . --no-build-vignettes
-    travis_wait 20 R CMD check projpred_*.tar.gz --ignore-vignettes --no-examples
+    travis_wait 30 R CMD check projpred_*.tar.gz --ignore-vignettes --no-examples
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -25,7 +25,8 @@ Imports:
     ggplot2, 
     Rcpp, 
     utils,
-    rngtools (>= 1.2.4)
+    rngtools (>= 1.2.4),
+    MASS
 LinkingTo: Rcpp, RcppArmadillo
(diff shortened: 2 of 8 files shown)


paul-buerkner:
Having the MASS dependency is fine since MASS is in the R system library anyway and thus natively shipped with R.

AlejandroCatalina:
Running `devtools::test()` locally passes all the tests. Anybody has a clue on why it seems to fail here?

paul-buerkner:
what happens if you run `devtools::check()`?

AlejandroCatalina:
There are 2 failures. I'll check them. The weird thing is that they don't come up if I run `devtools::test()`. And in anywise the output is still different from Travis.

AlejandroCatalina:
The problem I'm having is that I can't replicate the failures 😅 .

paul-buerkner:
I will check the non-gaussian branche if I get any failues and try to fix them.

AlejandroCatalina:
Let me know if you find a way to reproduce the failures. When I run the code that is supposed to fail it just works.

AlejandroCatalina:
I commented out the failing tests for the moment. The bug is weird because the failing code appears in other places and it doesn't fail there. Nonetheless, I think we can keep going for the moment. If you agree, @paul-buerkner , I'll merge this and open new PRs for subsequent work.

paul-buerkner:
Sounds good to me. Thanks!

Alejandro Catalina <notifications@github.com> schrieb am Mo., 6. Apr. 2020,
16:27:

PR #53, opened 2020-04-20
Deep refactor

AlejandroCatalina:
This is meant to be the last big PR before merging develop. This includes (growing list)

- [x] format and refactor code (superficial beautify).
- [x] clean `vsel` and `cvsel` structures and remove unneeded fields (do we need `pctch`?).
- [x] check redundant functions/code.
- [x] parse transformations in the response in formulas.
- [x] check `d_test` structure, do we need this? what should it contain?
- [x] rename `mle_helpers` to something more general.
- [x] make sure tests are passing after refactoring.


AlejandroCatalina:
I'm quite satisfied at this point but I'm sure there are more things to refactor.
@paul-buerkner I think this might be a good point for you to look at the codebase and point at things that you don't like or look strange (I'm looking at the code everyday and I'm so used to it that everything seems alright).

paul-buerkner:
Thank you! I will take a detailed look in the next couple of days.

paul-buerkner:
Are there any particular R files you want me to check?

AlejandroCatalina:
Not really, we're looking at the entire codebase. I feel that there's not too much left to polish in `cv_varsel.R` and `varsel.R` because I've worked quite some time on them. Potential files where we can polish more include:

- `project.R`
- `projfun.R`
- `summary_funs.R`
- `methods.R`

But in the end we are going over the whole codebase (I have worked on all files already but to a different degree in each of them). We can start with those potential files to narrow the list down.

paul-buerkner:
I see. Ok, I will try my best to check consistency and non-redundancy but I have not much in-depth knowledge of the projpred code base so I may miss some important things right now.

Anyway, I start with a few comments on the files you mentioned above.

- [ ] `number_samples` should be renamed to `ndraws` if the former refers to the number of posterior draws.
- [ ] `number_clusters´ should be renamed to `nclusters`
- [ ] `nv` should be renamed to `nvariables` or even `nterms` if it refers to included terms rather than individual variables.
- [ ] A lot of arguments are passed to functions by order rather than by name. The latter is safer and more explicit. There is no need for this practice for simple functions with 1-3 arguments or so but for more arguments, explicit argument names would be helpful.
- [ ] `project_submodel` and `.get_submodels` seem to have some redundancies. Can you check if this is necessary?
- [ ] In `methods.R` there seems to be some old code requiring new `x`, and `offset`. I had in mind that we changed the interface in regard to those argument, didn't we? In general it needs to be checked if the argument names there match our favored names, e.g., `ndraws` of the number of posterior draws.
- [ ] Check if the `as.matrix` and `t` methods in  `methods.R` are exported. Because if they are, they may cause a lot of trouble for users if they overwrite usual behavior.
- [ ] Indices should be written out or abbreviated as `_ids` not `ind` as in `cvind`.

AlejandroCatalina:
Yes, I'll address those! 

Regarding the interface of some of the functions in `methods.R`, you're right and I'll work on updating some arguments like `x` and `offset`. 

In some projection functions we accept `weightsnew` and `offsetnew` because we are working on the same refmodel but we may need different parameters for future data, so I'm not sure about how to handle those, I'll think on it and update the functions.

I will also update some arguments to named arguments.

PR #60, opened 2020-08-07
New projpred is coming

AlejandroCatalina:
Hi everyone, 

We are finally at a point where develop has been stable for quite some time with all tests passing and we might be good for an alpha release. We will also improve the documentation and make sure that everything is up to date, but that might come after we've merged this.

This is huge so we'll leave this open for some time so that everyone has the opportunity to comment.


AlejandroCatalina:
I'm on it. I believe the current version is already packaged? Otherwise, we can create a milestone right now on master. I took a look at the conflicts locally and they are related to doc as well, so I'll solve those issues you mention first.

Thank you for your time!

paul-buerkner:
Thank you! If the current version is already packaged, than this is fine. I think the current CRAN release is missing 2 non-essential commits so this should be fine.

AlejandroCatalina:
There are now only 2 warnings I believe. One of them refers to using `:::` to call an internal method of `mgcv` that we kind of need, I guess we can work around that by copying the method to projpred but that would be ugly? The second one refers to consistency of generic methods, and cites some examples as in:

```
   varsel:
     function(object, ...)
   varsel.refmodel:
     function(refmodel, d_test, method, ndraws, nclusters, ndraws_pred,
              nclusters_pred, cv_search, nterms_max, intercept, verbose,
              lambda_min_ratio, nlambda, thresh, regul, penalty,
              search_terms, ...)
```

Does this mean that I have to write the same arguments in the generic, or that I should have defaults for all new arguments in the specific methods?

paul-buerkner:
The use of internal function of other packages is forbidden by CRAN policies so that is a problem. Shall we have a call to discuss this issue and what to do about it? (I will write you on slack to find a time).

The problem with the generic - method mismatch is because in the generic, the first slot if called "object" while in the method it is called "refmodel". I suggest changing it to "object" in all of the methods, which should fix this problem.

AlejandroCatalina:
There is now only one consistency issue that has to do with `plot`. The warning says:

```
   plot:
     function(object)
   plot.vsel:
     function(object, nterms_max, stats, deltas, alpha, baseline, ...)
```

Can we do anything about this? We have a generic definition as:

```

#' Plot summary statistics related to variable selection
#' @inheritParams summary.vsel
#' @export
plot <- function(object, ...) {
  UseMethod("plot", object)
}
```

paul-buerkner:
plot is a base R generic and we should simply remove our own plot generic. It may be that we may need to add `#' @method plot vsel` if this is not automatically detected.

AlejandroCatalina:
Done, I'm doing a final local check.

vignettes/quickstart_glmm.Rmd
```diff
@@ -0,0 +1,354 @@
...
+  geom_vline(xintercept = as.numeric(y_subset), color = "red") +
+  xlab("y1_rep")
+```
+
+## Poisson example
+
+In this case we will download count data from the web,
+
+```{r}
+data_pois <- read.table("data_pois.csv", header = TRUE)
```
avehtari, line 154: line 154 is not downloading anything from the web as claimed on line 151

vignettes/quickstart_glmm.Rmd
```diff
@@ -0,0 +1,354 @@
...
+```
+
+These data correspond to a phylogenetic dataset where the phenotype is expressed
+as counts. This kind of data is relevant in evolutionary biology when data of
+many species are analyzed at the same time. The model we fit here is borrowed
+from *Modern Phylogenetic Comparative Methods and the application in
+Evolutionary Biology* (de Villemeruil & Nakagawa, 2014). The necessary data can
+also be downloaded from the corresponding website
+(http://www.mpcm-evolution.com/). The model is taken from the *Estimating
```
avehtari, line 163: Line 159 The model... 
Line 163 The model...
considering rewording one these sentences

vignettes/quickstart_glmm.Rmd
```diff
@@ -0,0 +1,354 @@
...
+
+As we did in the previous example, we can perform variable selection and look
+at the projection.
+ 
+```{r, cache=TRUE, results="hide", message=FALSE, warning=FALSE}
+vs <- varsel(fit)
+```
+
+Because our model includes group effects, the only available search method is
+forward search, which is a bit slower than L_1 search but is quite accurate.
```
avehtari, line 183: see the earlier comment

vignettes/quickstart_glmm.Rmd
```diff
@@ -0,0 +1,354 @@
...
+  geom_point(aes(x = pred$pred, y = y_mu)) +
+  geom_line(aes(x=xaxis, y = exp(xaxis)), color = "red") +
+  labs(x = "prediction", y = "y")
+```
+
+## Bernoulli example
+
+For both of the previous cases, running variable selection with the full data
+was actually enough because they were pretty simple. In this section we work on
+a slightly more complex data set.
```
avehtari, line 246: maybe change
complex -> difficult

vignettes/quickstart_glmm.Rmd
```diff
@@ -0,0 +1,354 @@
...
+   `ndraws_pred` helps but also increases the computational cost. Re fitting the
+   reference model ensuring a narrower posterior (usually employing a stronger
+   sparsifying prior) would have a similar effect. This is usually the case when
+   the difference in predictive performance is not very large.
+2. For non-Gaussian models the source of the inaccuracy may come from the fact
+   that we are running an approximate projection. In this case, increasing the
+   number of draws for the prediction `ndraws_pred` should also help.
+3. Given that `varsel` computes the expected predictive performance of the
+   reference model from the in-sample mean, it may sometimes result in
+   overconfident results. Running `cv_varsel` with LOO cross validation computes
```
avehtari, line 311: cross validation -> cross-validation

vignettes/quickstart_glmm.Rmd
```diff
@@ -0,0 +1,354 @@
...
+   the predictive performance by taking PSIS-LOO weights into account, usually
+   reporting a more realistic ELPD.
+
+By default, `cv_varsel` will run LOO cross validation, and will try to run a
```
avehtari, line 315: cross validation -> cross-validation

R/cv_varsel.R
```diff
@@ -1,46 +1,63 @@
...
-#' faster computation but higher uncertainty (larger errorbars) in the accuracy estimation.
-#' Default is to use all observations, but for faster experimentation, one can set this to a small value such as 100.
-#' Only applicable if \code{cv_method = 'LOO'}. 
-#' @param K Number of folds in the k-fold cross validation. Default is 5 for genuine
-#' reference models and 10 for datafits (that is, for penalized maximum likelihood estimation).
+#' @param verbose Whether to print out some information during the validation,
+#'   Default is TRUE.
+#' @param cv_method The cross-validation method, either 'LOO' or 'kfold'.
+#'   Default is 'LOO'.
+#' @param nloo Number of observations used to compute the LOO validation
```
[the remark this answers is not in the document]
AlejandroCatalina: I'll note this in our TODO list.

R/methods.R
```diff
@@ -468,166 +563,269 @@ print.cvsel <- function(x, digits=2, ...) {
...
+##' Divide indices from 1 to \code{n} into subsets for \code{k}-fold cross
+##' validation. These functions are potentially useful when creating the
+##' \code{cvfits} and \code{cvfun} arguments for
+##' \link[=init_refmodel]{init_refmodel}. The returned value is different for
+##' these two methods, see below for details.
+##'
+##' @name cv-indices
+##'
+##' @param n Number of data points.
+##' @param k Number of folds. Must be at least 2 and not exceed \code{n}.
```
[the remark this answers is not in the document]
AlejandroCatalina: renamed to K

R/project.R
```diff
@@ -1,42 +1,55 @@
...
-#' @param ns Number of samples to be projected. Ignored if \code{nc} is specified. Default is 400.
-#' @param nc Number of clusters in the clustered projection.
+#'
+#' @name project
+#'
+#' @param object Either a \code{refmodel}-type object created by
+#'   \link[=get_refmodel]{get_refmodel} or \link[=init_refmodel]{init_refmodel},
+#'   or an object which can be converted to a reference model using
+#'   \link[=get_refmodel]{get_refmodel}.
+#' @param nterms Number of variables in the submodel (the variable combination
```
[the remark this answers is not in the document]
AlejandroCatalina: That's a good point. There is a difference in fact, as we used in the paper, variables would correspond to individual variables in the data, whereas terms can correspond to a number of columns (splines expansion, multilevel terms, etc). Posting here for reference. I'll change it to terms here and in other locations.

R/projpred-package.R
```diff
@@ -13,9 +13,10 @@
 #' 
 #' @description Description
 #' 
-#' \pkg{projpred} is an R package to perform projection predictive variable selection for generalized
-#' linear models. The package is aimed to be compatible with \pkg{rstanarm} but also other
-#' reference models can be used (see function \code{\link{init_refmodel}}). 
+#' \pkg{projpred} is an R package to perform projection predictive variable (feature)
+#' selection for generalized linear models. The package is aimed to be compatible with
```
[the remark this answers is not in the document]
AlejandroCatalina: This PR includes GLMMs and GAMMs indeed, I'll update the description!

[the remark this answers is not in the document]
AlejandroCatalina: At this point, we should also add a GAMM vignette when possible!

AlejandroCatalina:
Should be good now @avehtari

jpiironen:
Hi, and sorry for a long delay. I now installed the dev-version on my laptop and started running the quickstart-vignette but I did not get far before I got this error:

```
Quitting from lines 103-104 (quickstart.Rmd) 
Error in init_refmodel(object, data, y, formula, family, ref_predfun,  : 
  argument "y" is missing, with no default
Calls: <Anonymous> ... get_refmodel.default -> init_refmodel -> extract_terms_response -> terms
In addition: Warning message:
Bulk Effective Samples Size (ESS) is too low, indicating posterior me'ans and medians may be unreliable.
Running the chains for more iterations may help. See
http://mc-stan.org/misc/warnings.html#bulk-ess 
Execution halted
```

Does it run fine for everybody else?

AlejandroCatalina:
I’ll take a quick look. There must be something out of scope because both Paul and I run both vignettes! Thanks.

Best,
Alejandro

AlejandroCatalina:
Oh, I see now. @jpiironen what seems to happen is that you need to install `brms@projpred` branch for `get_refmodel` to work on brmsfit models. That's something we need to be very explicit about to avoid these cases.

jpiironen:
Thanks @AlejandroCatalina, I got everything running. I briefly went through the two vignettes, I don't have any major concerns but just a few comments/questions:

- From quickstart-vignette: there's comment in one code block `# plot predictive performance on training data` although this actually plots the result of `cv_varsel`

- Plotting the above result looks somewhat different than earlier (if you look at https://mc-stan.org/projpred/articles/quickstart.html for example). I wonder what's the reason? Is the L1-penalized search still using the same C++ code, is is it nowadays done differently?

- At the end of the vignette, there's an explanation: "Here's an example prediction for the first data point", but if you look at the code it picks one data point randomly

- Was there a reason to remove the example for the usage of custom reference models? I guess that would be useful for some people. Or is this something we do not support anymore?


- Type `?varsel` and look a the Value-section, it says: 
```
An object of type vsel that contains information about the feature selection. The fields are not #' meant to be accessed directly by the user but instead via the helper #' functions (see the vignettes or type ?projpred #' to see the main functions in the package.)
```
  So it has these extra #' -markers there (typos; should be removed). Would be good to check other docs as well for similar things.


- In vignette quickstart_glmm there's data loaded from 'data_pois.csv', I guess it could be put into a similar file as df_gaussian in quickstart-vignette for consistency. Maybe it could also be some actual name, since it seems to be a real world data set.

jpiironen:
(Clicked "Close with comment" by accident)

AlejandroCatalina:
Thanks for the comments, I’ll address the typos (!!) and answer your questions:

1. The C++ code is exactly the same, the l1 search happens the same but there might be a change in how we compute the final projections, because we use pure maximum likelihood estimates under the hood instead of ridge regression with a small penalisation constant. That could be a source for minimal discrepancies. Would that maybe answer the difference?
2. The custom reference models vignette is going to be redone because the way of initialising a reference model is quite different now, so it needs to be completely redone. It is indeed very interesting to have for many people so it’s definitely up on the todo list (just a bit busy with other things at the moment).
3. I’ll put data_pois.csv in a similar format as df_gaussian with a proper name, thanks!


Best,
Alejandro

jpiironen:
Thanks Alejandro. 

1. This is interesting. The default amount of extra penalization/regularization was very small, and especially in the Gaussian projection it hardly made any difference whether you set it to zero or used the default. Therefore I can't believe that would explain such a big difference that I now seeprojpred_old
projpred_new

AlejandroCatalina:
Yes, that’s interesting. I believe this will come down to how lme4 is computing the MLE, because that’s the only difference if recall correctly. I will take a look nonetheless! Thanks!

Best,
Alejandro

jpiironen:
Ok, great. Once you have the vignette built with the latest changes, can you post a figure about what the result looks like? So I can visually compare it to the old one.

AlejandroCatalina:
Here it is, for reference
image

jpiironen:
Oh yeah, now looks much better! I couldn't spot a difference to the old one.

AlejandroCatalina:
Great! Waiting for @avehtari then to approve before going forward! Thanks for the time put into this.

paul-buerkner:
I will now merge this PR. Thank you @AlejandroCatalina for your work on improving projpred! This is a huge jump in featured models and modularization of the code!

Issue #72, opened 2021-02-08
Incorrect column names in as.matrix.projection()


fweber144:
It seems like the column names of the matrix returned by `as.matrix.projection()` are not correct (at least not always). Here is a reprex: First run
```r
# Original source: https://github.com/stan-dev/projpred/blob/master/README.md
library(projpred)
library(rstanarm)
options(mc.cores = parallel::detectCores(logical = FALSE))
data("df_gaussian", package = "projpred")
df_gaussian <- df_gaussian[1:40, ]
mydat <- cbind("y" = df_gaussian$y, as.data.frame(df_gaussian$x))
mydat$mygroup <- gl(n = 8, k = floor(nrow(mydat) / 8),
                    labels = paste0("gr", seq_len(8)))
set.seed(457211)
mydat$noise <- rnorm(nrow(mydat))
mygroup_icpts_truth <- rnorm(nlevels(mydat$mygroup), sd = 6)
mygroup_V1_truth <- rnorm(nlevels(mydat$mygroup), sd = 6)
mydat$y <- mydat$y + mygroup_icpts_truth[as.numeric(mydat$mygroup)]
mydat$y <- mydat$y + mygroup_V1_truth[as.numeric(mydat$mygroup)] * mydat$V1

myfit <- stan_glmer(y ~ V1 + V2 + V3 + V4 + V5 + noise + (1 + V1 | mygroup),
                    data = mydat,
                    seed = 1140350788)

cvs <- cv_varsel(myfit)
myproj <- project(cvs,
                  nterms = length(solution_terms(cvs)),
                  ndraws = ceiling(0.05 * nrow(as.matrix(myfit))))
debugonce(projpred:::as.matrix.projection)
myproj_mat <- as.matrix(myproj)
```
and then debug the last call up to (but excluding) line
https://github.com/stan-dev/projpred/blob/04326a1ceaedccf4883a1a6fb6c04b77d186c08d/R/methods.R#L835
If you then compare `x$solution_terms` and `colnames(res)`, then you should see that the ordering of the terms in `x$solution_terms` doesn't match the ordering in `colnames(res)`:
```r
# x$solution_terms:
# [1] "(1 | mygroup)"  "V1"             "(V1 | mygroup)" "V3"             "V5"             "noise"          "V4"            
# [8] "V2"  
# colnames(res):
# [1] "(Intercept)"         "V1"                  "V3"                  "V5"                  "noise"              
[...]

PR #73, opened 2021-02-08
Fix incorrect column names in as.matrix.projection()

fweber144:
This should fix issue #72. Of course, you don't have to use my development-specific version number.


fweber144:
Instead of my fix from commit ec0224d963c55e59fb4f3e219846ef67339a180e, one could also consider replacing
https://github.com/stan-dev/projpred/blob/04326a1ceaedccf4883a1a6fb6c04b77d186c08d/R/methods.R#L830-L836
(or probably even
https://github.com/stan-dev/projpred/blob/04326a1ceaedccf4883a1a6fb6c04b77d186c08d/R/methods.R#L829-L837
) by
```r
colnames(res) <- gsub("^1|^alpha|\\(Intercept\\)", "Intercept", colnames(res))
```
but I haven't tested that yet.

fweber144:
Closed since PR #75 (merged) includes this PR.

PR #74, opened 2021-02-08
as.matrix.lmerMod(): Extract variance components, not empirical SD

fweber144:
In `as.matrix.lmerMod()`, I think it's better to extract the variance components, not the empirical standard deviation (SD) of the random effects. For example, Gelman (2005) discusses the difference between the interpretation of the empirical and the population SD. Furthermore, with "variance components", I don't only mean the population SD, but also correlations in case of multiple terms per group-level term (e.g. `(1 + x1 | group)`). The variance components may be extracted by `lme4::VarCorr()`. All the other code from commit 9618d735664fedc76c2eceb2bbe169b0c03cbbe4 is just for adding meaningful names.

This is the first of the two PRs mentioned [here](https://github.com/stan-dev/projpred/issues/72#issuecomment-775073670). Because of the possibly present correlation parameters, this PR builds upon PR #73 to ensure that `as.matrix.projection()` handles the column names of its output matrix correctly.

**References:**
Gelman, A. (2005). Analysis of variance—Why it is more important than ever. *The Annals of Statistics*, 33(1), 1–53. https://doi.org/10.1214/009053604000001048

**EDITS:**
* [Here](https://github.com/fweber144/projpred/compare/fweber144:proj_mat_fix..fweber144:proj_mat_varcomp) is the diff for this PR and PR #73.
* The empirical SD may still be calculated manually be the user if `as.matrix.lmerMod()` not only returns the variance components, but also the estimated random effects themselves (as given by `lme4::ranef()`). This is what PR #75 does.


AlejandroCatalina:
Can we do these PRs against develop branch instead of master?

fweber144:
> Can we do these PRs against develop branch instead of master?

Yes, I can do that. I don't have the time right now, but I'll do that asap.

fweber144:
> > Can we do these PRs against develop branch instead of master?
> 
> Yes, I can do that. I don't have the time right now, but I'll do that asap.

Done now.

fweber144:
Closed since PR #75 (merged) includes this PR.

PR #75, opened 2021-02-08
as.matrix.lmerMod(): Extract random effects

fweber144:
It would be nice if `as.matrix.lmerMod()` would also extract the (estimated) random effects themselves. Or are there any theoretical arguments against that? If not, then the code from this PR could perform that extraction. Note that the essential call is that to `lme4::ranef()`; the code from commit 5fa6d3f792546ee38d72fcb67ba473369f3559c3 is just for adding meaningful names.

This is the second of the two PRs mentioned [here](https://github.com/stan-dev/projpred/issues/72#issuecomment-775073670). Because of the multiple parameters (multiple random effects) per group-level term, this PR builds upon PR #73 to ensure that `as.matrix.projection()` handles the column names of its output matrix correctly.

Furthermore, this PR builds upon PR #74 since PR #74 is enhanced by this PR: If `as.matrix.lmerMod()` returns the random effects themselves, then the user may calculate their empirical SD manually.

**EDITS:**
* [Here](https://github.com/fweber144/projpred/compare/fweber144:proj_mat_fix..fweber144:proj_mat_ranef) is the diff for this PR and PR #73.
* [Here](https://github.com/fweber144/projpred/compare/fweber144:proj_mat_varcomp..fweber144:proj_mat_ranef) is the diff for this PR and PR #74.

PR #82, opened 2021-03-03
as.matrix.lmerMod(): Use brms's naming convention

fweber144:
As discussed in #75, it probably makes sense to use brms's convention for parameter names in the matrix returned by `as.matrix.projection()`. With this PR, I implemented this for GLMMs (i.e. I modified `as.matrix.lmerMod()`). Other models (GLMs, GAMs, GAMMs; that's it, I guess) still need to be tackled, though.

In commit 51eba652aa2239e59f83a59e2abe5e60329b7ab8, I stripped trailing horizontal whitespace for the whole "methods.R" file since that seems to be projpred's convention. To make this clear for new contributors (assuming they use RStudio), versioning the "projpred.Rproj" file (with `StripTrailingWhitespace: Yes`) might be a good idea. See e.g. [here](https://community.rstudio.com/t/should-rproj-files-be-added-to-gitignore/1269) for a discussion.

### DIFF
--- a/R/methods.R
+++ b/R/methods.R
@@ -61,10 +61,10 @@
 #'   x <- matrix(rnorm(n*d), nrow=n)
 #'   y <- x[,1] + 0.5*rnorm(n)
 #'   data <- data.frame(x,y)
-#'   
+#'
 #'   fit <- rstanarm::stan_glm(y ~ X1 + X2 + X3 + X4 + X5, gaussian(), data=data, chains=2, iter=500)
 #'   vs <- varsel(fit)
-#'   
+#'
 #'   # compute predictions with 4 variables at the training points
 #'   pred <- proj_linpred(vs, newdata = data, nv = 4)
 #'   pred <- proj_predict(vs, newdata = data, nv = 4)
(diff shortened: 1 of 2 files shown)


AlejandroCatalina:
I think it may make sense to write a general method to handle variable names that defers to specific cases when it encounters specific terms, for instance multilevel terms and so on, but the baseline method can be shared. I’ll take a look at your implementation and comment further.

Best,
Alejandro

R/methods.R
```diff
@@ -770,8 +770,18 @@ as.matrix.glm <- function(x, ...) {
...
   population_effects <- lme4::fixef(x)
+
+  # Use brms's naming convention:
+  names(population_effects) <- gsub(
+    "\\(Intercept\\)",
+    "Intercept",
+    names(population_effects)
+  )
+  names(population_effects) <- paste0("b_", names(population_effects))
+
```
AlejandroCatalina, line 781: We can probably put this under `as.matrix.(g)lm` and call it here.

AlejandroCatalina:
This looks okay, I'm just thinking about the redundancy in the final part of the code to set the names for group_vc_raw and the group effects themselves. It seems some code can be abstracted but I'm not sure of the best approach. I'll play a bit locally and see if my abstraction looks any better.

AlejandroCatalina:
I did the merge locally already and added couple things.

PR #92, opened 2021-03-19
Miscellaneous fixes and improvements

fweber144:
I collected some minor no-op fixes (see **EDIT** below), in particular:

* Added an .Rproj file to ensure common settings for people forking projpred (e.g. stripping trailing horizontal whitespace and enabling roxygen2 documentation).
* Added file "/src/projpred.dll" to the Git ignore list (created when compiling the package's C++ code on Windows).
* Added some older changes to the NEWS file.
* Several improvements in the documentation.

For the latter point, note that I have finished working (so far) on the documentation of varsel.R, but only started working on the documentation of project.R. Similarly to my documentation changes in varsel.R, we could also adopt the documentation of project.R and all other files. But I first wanted to hear from you if all those `\code{}` additions are ok.

When improving the formatting of the whole documentation, one could also think about switching to Markdown syntax (supported by roxygen2 version >= 6.0.0; see e.g. [here](https://cran.r-project.org/web/packages/roxygen2/vignettes/rd-formatting.html) for details) which makes things a lot simpler. For example, `\link[=<function_name>]{\code{<function_name>}}` is not possible in Rd syntax, so all those `\link[=<function_name>]{<function_name>}` (= `\link{<function_name>}`) occurrences would have to be converted to `\code{\link{<function_name>}}` (or better also appending parentheses: `\code{\link[=<function_name>]{<function_name>()}}`). In Markdown syntax, one would simply write `[<function_name>()]`.

(Another note: First, I increased the development-specific version number, but reverted that later since I realized that different development-specific version numbers are probably too fine-grained.)

**EDIT:** In the meantime, the changes have become larger and there is also at least one fix which is *not* a no-op one. I'll always try to indicate when a change is *not* a no-op one.


fweber144:
Just pushed another commit (4dfb4b7).

fweber144:
Commit d9fea25a649f1290643415e44807ccceec1da20e is not a no-op fix. Do you want me to rename this PR?

AlejandroCatalina:
It has grown a bit and it includes various things, you can rename it to something that fits better?

Best,
Alejandro

fweber144:
> It has grown a bit and it includes various things, you can rename it to something that fits better?

Sure. I hope it's ok if I only change the GitHub name, not the branch name.

fweber144:
Just FYI: The reason for commit 616612b was not only for improving debugging, but also because of scoping: In R, a function looks for an object in its parent frame, as shown below. And since the auxiliary function (named `proj_predict_local()` in the example below) was defined in `proj_linpred()` (or `proj_predict()`), changing the weights or the offsets on the way downstream to their actual use would have had no effect. As far as I can tell, this scoping peculiarity did not seem to have undesired effects in this particular case (since the weights and the offsets don't seem to be modified on the downstream way) but it could have had undesired effects in the future.

This small example mimics the scoping in this particular case:
```r
proj_helper <- function(proj_predict_local, x) {
  x <- x / 2
  y <- "dummy"
  proj_predict_local(y)
}

proj_linpred <- function(x) {
  proj_predict_local <- function(y) {
    # Do something with y (not important here) and then use x:
    x + 1
  }
  proj_helper(proj_predict_local, x)
}

proj_linpred(3)
## --> Gives 3 + 1 = 4, not 3 / 2 + 1 = 2.5
```

AlejandroCatalina:
Sure. Ping me when the PR is ready and I’ll take a look.

Best,
Alejandro

fweber144:
Yes, that's probably the right way to go. Unfortunately, I guess you'll have to iterate through all of the commits in turn and decide which ones you want and which ones not. Sorry for this tedious task. Tell me if you need information at specific places.

Anyhow, I think there are a few things I would still like to check and perhaps modify. I'll ping you as soon as it's ready.

fweber144:
Note: Commit 104e633be16763bd4b03dffaf4dbb648e2799156 is *not* a no-op fix and commits c1f4c17, 0502770, af8bd7a also not really.

fweber144:
Notes:
* I have included some minor changes for `ndraws` and `nclusters` and also applied the new behavior to `project()`.
* Commits 6b80f47, 4ef95d3, and ce7d1e0 are *not* no-op fixes.

fweber144:
@AlejandroCatalina: This PR is ready to be merged now.

fweber144:
If you have the time, it would be nice to hear from you what you think about my questions and comments [above](https://github.com/stan-dev/projpred/pull/92#issue-596399046). For the `.Rproj` thing, see also [this comment](https://github.com/stan-dev/projpred/pull/82#issue-584231523).

AlejandroCatalina:
Great, I'll take a look at the changes now. 

Regarding your questions:
- I'm okay with changing the documentation to Markdown if it makes things simpler. We largely inherited the previous version of the documentation but I'm aware there is a lot to improve in that department. 
- I don't use Studio myself so I don't have a strong opinion regarding `.Rproj` and its properties, but it seems like including it in the git repo makes sense to me.

R/cv_varsel.R
```diff
@@ -190,20 +201,20 @@ cv_varsel.refmodel <- function(object, method = NULL, cv_method = NULL,
...
-      cv_search = cv_search, nterms_max = nterms_max - 1,
-      intercept = intercept, penalty = penalty, verbose = verbose,
-      lambda_min_ratio = lambda_min_ratio, nlambda = nlambda, regul = regul,
-      search_terms = search_terms, seed = seed
+                  method = method, ndraws = ndraws, nclusters = nclusters,
+                  ndraws_pred = ndraws_pred, nclusters_pred = nclusters_pred,
+                  cv_search = cv_search, nterms_max = nterms_max - 1,
+                  intercept = intercept, penalty = penalty, verbose = verbose,
+                  lambda_min_ratio = lambda_min_ratio, nlambda = nlambda,
+                  regul = regul, search_terms = search_terms, seed = seed
```
AlejandroCatalina, line 209: Is this indentation an R convention?

R/methods.R
```diff
@@ -81,58 +96,61 @@ NULL
...
+    ## projs is some other object, not containing an element called "family" (so
+    ## it could be a `proj_list` but must not necessarily)
+    if (any(sapply(projs, function(x) !("family" %in% names(x))))) {
+      stop("Invalid object supplied to argument `object`.")
     }
   }
 
   if (is.null(newdata)) {
     ## pick first projection's function
-    newdata <- proj[[1]]$refmodel$fetch_data()
```
AlejandroCatalina, line 108: I'm wondering, can we create a new structure `projection_list` that has all common slots as attributes? That seems a way better way of not duplicating all the internal structures. We should anyway make sure that when using any of the individual projections we have all the required slots available. I'll think on this.

fweber144:
> I'm okay with changing the documentation to Markdown if it makes things simpler. We largely inherited the previous version of the documentation but I'm aware there is a lot to improve in that department.

Alright, then I'll perform the switch to Markdown syntax as soon as I have the time (but it doesn't have top priority, I guess). I'll create a separate PR for that, though.

R/cv_varsel.R
fweber144: Not specifically, I think. As you know, setting a specific margin width (80 in projpred, if I saw that correctly) is a general computer science convention. And for the indentation, I follow RStudio's automatic indentation as well as [this style guide](http://adv-r.had.co.nz/Style.html). Commits 4c23b44 to 4ffe4f5 were just performed to make life easier when working in RStudio: I regularly use `Ctrl + A` and `Ctrl + I` to automatically indent the whole code of the current file. Lines which do not follow RStudio's automatic indentation guide yet are then re-indented which causes undesired changes which are not related to the piece of code I am currently working on. So I thought it would be a good idea to re-indent all R files here in this PR so that in the future, re-indenting in RStudio doesn't produce undesired changes which are not related to the piece of code I (or other RStudio users) am/are currently working on.

R/methods.R
fweber144: Yes, probably a good idea. Or simply always force a list (which is of length one in the case of a single submodel).

fweber144: In any case, that's a larger change, I guess. So perhaps it should be performed outside of this PR.

R/cv_varsel.R
AlejandroCatalina: Yes, I enforce 80 characters per line but my indentation system is clearly different, which messes up with my own rules. I'll take a look but I'll most likely accept this one because it seems most people work with RStudio.

fweber144:
> When improving the formatting of the whole documentation, one could also think about switching to Markdown syntax [...]

Solved by PRs #226 and #233.

Issue #97, opened 2021-03-22
proj_linpred() requires newdata


fweber144:
Contrary to section "Value" in its documentation, it seems like `proj_linpred()` requires argument `newdata` to be specified:
```r
library(projpred)
library(rstanarm)
options(mc.cores = parallel::detectCores(logical = FALSE))
data('df_gaussian', package = 'projpred')
mydat <- cbind("y" = df_gaussian$y, as.data.frame(df_gaussian$x))
myfit <- stan_glm(y ~ V1 + V2 + V3 + V4 + V5,
                  family = gaussian(),
                  data = mydat,
                  prior = hs(df = 1, global_scale=0.01), seed = 1140350788)

# Does not work:
mylinpred <- proj_linpred(myfit,
                          solution_terms = c("V3", "V2"),
                          ndraws = 20)
## --> Gives:
## Error in proj_helper(object = object, newdata = newdata, offsetnew = offsetnew,  :
##   argument "newdata" is missing, with no default

# Works:
mylinpred <- proj_linpred(myfit,
                          solution_terms = c("V3", "V2"),
                          newdata = mydat,
                          ndraws = 20)

```

fweber144:
Thanks for addressing this in commit 2207f4397306a387f58285068d9781161f6c9542. However, when I try to run the reprex from above, it still doesn't work. One potential issue is that the `proj[[1]]$extract_model_data()` call returns a list here. I managed to fix that (PR upcoming) but another issue seems to be that the `proj[[1]]$extract_model_data()` call does not return the data's columns (only `weights` and `offset`).

AlejandroCatalina:
As we are predicting for future data, the target is not guaranteed to be there. We only use that function to extract weights and offset and then pass the data frame again for prediction. I think I made a mistake overriding data frame! I’ll take a look at your PR and make any necessary fixes.

Best,
Alejandro

PR #98, opened 2021-03-24
Continue fixing #97

fweber144:
Discussed in #97 (more precisely: [this comment](https://github.com/stan-dev/projpred/issues/97#issuecomment-805639232)).


AlejandroCatalina:
So, I took a look at this and I realised I made a mistake. What we need is simply to fetch the reference model's training data so this PR is not needed now. I have fixed that locally now. Sorry about that 😅 .

fweber144:
No problem :)

Issue #96, opened 2021-03-19
Omitting intercept in submodel(s) does not work


fweber144:
It seems like setting argument `intercept = FALSE` in `project()` has no effect, as shown by the reprex below (tested on develop branch). The problem might be that `intercept` is not taken into account in these lines: https://github.com/stan-dev/projpred/blob/0cabd73548eec32e9b675f15cd39d377a40f180a/R/projfun.R#L84-L92 but the problem might also be that `project_submodel()` has an argument `intercept` which is not used, though.

In the reprex below, I illustrate the issue with `project()` but I think it also occurs when projecting in `varsel()` and `cv_varsel()`.

```r
library(projpred)
library(rstanarm)
options(mc.cores = parallel::detectCores(logical = FALSE))
data('df_gaussian', package = 'projpred')
mydat <- cbind("y" = df_gaussian$y, as.data.frame(df_gaussian$x))
myfit <- stan_glm(y ~ V1 + V2 + V3 + V4 + V5,
                  family = gaussian(),
                  data = mydat,
                  prior = hs(df = 1, global_scale=0.01), seed = 1140350788)

myproj <- project(myfit,
                  solution_terms = c("V3", "V2"),
                  ndraws = 20,
                  intercept = FALSE)
colnames(as.matrix(myproj))
## --> Gives:
## [1] "Intercept" "V3"        "V2"        "sigma"
```

AlejandroCatalina:
When discussing this, we got to the conclusion that we would not allow projections without intercept. Therefore, I'm removing this argument altogether, as we don't think it makes sense to compute projections without intercept in general.

fweber144:
Thanks for addressing this in commit 55cf78bc81ae92c1fc9dd041b67bead231552bfc. As far as I understood that commit, you're using `refmodel$intercept` as a guidance on whether to include the intercept in the submodels or not. This is a good solution, I think. (In contrast, forcing the submodels to always include an intercept would probably have been less desirable, as there may be cases like [this one](https://discourse.mc-stan.org/t/variable-selection-with-ordinal-model/17242) where excluding the intercept from the submodels indeed makes sense. But I agree that these are probably the same cases where the intercept would also be excluded from the reference model, so using `refmodel$intercept` as a guidance should be fine.)

However, if I now run the reprex from above but exclude the intercept from the reference model, then the submodel still contains an intercept (and that intercept is non-zero):
```r
library(projpred)
library(rstanarm)
[...]

PR #100, opened 2021-03-26
Fixes for no-intercept models

fweber144:
This is related to #96 and tries to fix issues occurring for reference models which do not have an intercept (so that the intercept has to be excluded from the projected submodels, too). Notes:

* I have only looked at the reprex from #96 which uses `project()`. In particular, I haven't looked at the projection(s) happening in `varsel()` and `cv_varsel()` calls for no-intercept reference models. I hope `varsel()` and `cv_varsel()` should also work now appropriately for that case, but I'm not 100% sure. Better check this.
* I haven't looked at the C++ code from `glm_ridge_c()` to see whether the intercept is properly taken into account there at all places where it needs to. But the fact that it has an argument `intercept` and that it returns an intercept of 0 for the reprex from #96 makes me hope that it should. However, you should probably check that.
* Currently, for no-intercept reference models, the object returned by `as.matrix.projection()` still contains a column for the intercept, but it's only zeros. Perhaps (for user-friendliness) it might be better to omit this zero-column.


AlejandroCatalina:
This should translate directly to `varsel()` and `cv_varsel()`, but it's best to add proper tests to `test_varsel.R` and `test_project.R`. Can you do that? Namely, fit `varsel` and `cv_varsel` both with and without intercepts and check that the submodels act accordingly, then project those `vsel` objects and check that intercept works as intended.

fweber144:
Yep, I'll do that.

fweber144:
Ok, this whole thing is more complex than I thought. First of all, my PR broke existing code for models which *do* have an intercept (so the majority of models). I'm sorry for that; I forgot to run the unit tests from **testthat**. I just pushed some commits which should fix at least that. Now I'll still have to adopt `varsel()` and `cv_varsel()` to no-intercept models. But that might take me some time since it requires more changes than I would have thought.

fweber144:
I think I'll be stopping here for now. As stated in the message of the last commit I just pushed, this PR probably needs more adjustments. The error messages I added in that last commit account for the unfinished state of this PR. If this PR doesn't get merged into `develop`, then I think at least the error messages should be added to `develop` (so that [this reprex](https://github.com/stan-dev/projpred/issues/96#issuecomment-808375740) throws an error).

If I get the time, I can perhaps continue working on this, but for now I'm lacking that time.

AlejandroCatalina:
Closing for now with #124 .

PR #110, opened 2021-04-07
Fixes for argument `nterms` of `proj_helper()`

fweber144:
See the commit messages for details.


AlejandroCatalina:
This one is failing some tests for some reason. Can you take a look at that?

fweber144:
Sure

fweber144:
I will have to revise this.

fweber144:
Now it should be ok. Sorry it took me a while to fix this.

fweber144:
Just FYI: The original idea of argument `nterms` seems to have been to "filter" `object` if it already results from `project()`. That wasn't possible, though (since `list(...)$nterms` was always `NULL`). This PR does *not* try to add a new argument for this "filtering" task. This is done in PR #92.

AlejandroCatalina:
Yeah, it happened that before the refactoring projections were handled differently. We introduced now the solution_terms argument in project that simplifies things, but earlier you could only pass integers and sometimes you wanted to predict for projections that were not done.

Best,
Alejandro

PR #113, opened 2021-04-08
Update NEWS file and bump version number

fweber144:
See the commit messages. A tag such as `v2.0.5` might be placed at commit ed55f74.


fweber144:
Now good to go

PR #116, opened 2021-04-08
vselsearch: Add `fit_elapsed_time` slot to `refmodel`

hsbadr:
The reference model of class `brmsfit` provides `fit$fit` instead of `fit$stanfit`.

This fixes the following error from `rstan::get_elapsed_time`:
```r
Error in .local(object, ...) :
  trying to get slot "mode" from an object of a basic class ("NULL") with no slots
```

**Example:** Let's use [the workflow vignette](https://htmlpreview.github.io/?https://raw.githubusercontent.com/stan-dev/projpred/workflow/vignettes/workflow.html):
```r
library(brms)
library(projpred)
library(dplyr)
library(ggplot2)
library(bayesplot)
theme_set(theme_classic())
#options(mc.cores = parallel::detectCores())

data('df_gaussian', package = 'projpred')

split_structure <- break_up_matrix_term(y ~ x, data = df_gaussian)
df_gaussian <- split_structure$data
formula <- split_structure$formula
d <- df_gaussian
n <- nrow(df_gaussian) # 100
D <- ncol(df_gaussian[, -1]) # 20
p0 <- 5 # prior guess for the number of relevant variables
tau0 <- p0/(D-p0) * 1/sqrt(n) # scale for tau (notice that stan_glm will automatically scale this by sigma)
fit <- brm(formula, family=gaussian(), data=df_gaussian,
           prior=prior(horseshoe(scale_global = tau0, scale_slab = 1), class=b),
           ## To make this vignette build fast, we use only 2 chains and
           ## 500 iterations. In practice, at least 4 chains should be 
           ## used and 2000 iterations might be required for reliable
           ## inference.
           seed=1, chains=2, iter=500)

search <- varsel_search(fit)
print(search)
```
```r
Error in .local(object, ...) :
  trying to get slot "mode" from an object of a basic class ("NULL") with no slots
```
```r
> traceback()
7: .local(object, ...)
6: rstan::get_elapsed_time(x$fit$stanfit) at stanfit-class.R#363
5: rstan::get_elapsed_time(x$fit$stanfit) at workflow.R#145
4: print.vselsearchsummary(summary)
3: print(summary) at workflow.R#106
2: print.vselsearch(search)
1: print(search)
```


AlejandroCatalina:
I am aware of this problem. The correct way of doing this is incorporating a new slot in `refmodel`. When we call `get_refmodel` each package should return a new value for `elapsed_time` that contains the elapsed time for the model fitting. We don't want to have hard coded switches in the code. 

You can incorporate this change currently for `get_refmodel.stanreg` and the necessary changes in `init_refmodel` if you are willing. Otherwise I recommend you do `fit$stanfit <- fit$fit` before calling `varsel_search` as this is a largely experimental branch and I doubt brms incorporates this change soon enough.

hsbadr:
Thanks @AlejandroCatalina! That makes sense to me. I though of this already, but wanted to push a quick fix. I agree that the `refmodel` object should be standard regardless of the Stan interface or how i was created. Let me try to update this as you suggested.

AlejandroCatalina:
You can of course ask anything you need, I’m thrilled you want to contribute to projpred! To address the case of `brmsfit` object you can have the switch in `init_refmodel`, which is not optimal but would work as a temporary solution.

Best,
Alejandro

hsbadr:
> To address the case of `brmsfit` object you can have the switch in `init_refmodel`, which is not optimal but would work as a temporary solution.

Yes, that's exactly what I'm going to do now.

AlejandroCatalina:
And then including a slot `refmodel$fit_elapsed_time` would allow for a general implementation of `print.vselsearchsummary`.

Best,
Alejandro

hsbadr:
@AlejandroCatalina On another note, I think the hard-coded parallel plan is not optimal. I'd prefer to let the user setup their own `future` and we use it internally, defaulting to `sequential` processing with a message. This allows using all the `future` backends, including running on a HPC cluster using `SLURM` for example.

AlejandroCatalina:
For sure, totally agreed. This is a very quick prototype we needed to speed up some tests we are running locally that are helping to polish the branch feature-wise. If you want to take a look at that we can coordinate the implementation.

Best,
Alejandro

hsbadr:
@AlejandroCatalina I think this is good to go.

hsbadr:
> For sure, totally agreed. This is a very quick prototype we needed to speed up some tests we are running locally that are helping to polish the branch feature-wise. If you want to take a look at that we can coordinate the implementation.

Sure. I'd like to contribute to the parallelism implementation if we can coordinate. It seems that you're working on it locally; so, I prefer to know your plan and join forces to avoid duplication of efforts.

AlejandroCatalina:
Well, my initial idea is to let the user set up the future backend and assume it has been registered already when the code runs. So we would directly run a foreach loop with dopar or dorng, more similar to what we do in cv_loo. Let me know if you think I’m missing something.

Best,
Alejandro

hsbadr:
> Well, my initial idea is to let the user set up the future backend and assume it has been registered already when the code runs. So we would directly run a foreach loop with dopar or dorng, more similar to what we do in cv_loo.

Sounds good to me. We may add a startup message to instruct the users if they're interested in parallel processing.

AlejandroCatalina:
Yes. I think the only necessary thing would be to remove the `doFutureRegister()` call from the file, and the various `future::plan(future::multicore, ..)` in case the user decides another backend.

Foreach should work for all of them, I’m using doFuture because it automatically exports needed objects. What are your thoughts on the whole scheme?

Best,
Alejandro

hsbadr:
> Yes. I think the only necessary thing would be to remove the `doFutureRegister()` call from the file, and the various `future::plan(future::multicore, ..)` in case the user decides another backend. Foreach should work for all of them, I’m using doFuture because it automatically exports needed objects. What are your thoughts on the whole scheme? Best, Alejandro

I use `DoFuture` in general unless it causes issues with `R-devel` or memory leaks with forked processes (`multicore`). The fallback is to create a cluster using `cl <- makeCluster(<# cores>)` and register the parallel `foreach` backend with `registerDoParallel(cl)`. For shared-memory parallelism, `multicore` is more efficient with `gc = TRUE, earlySignal = TRUE` but it isn't supported on Windows. Other `future` options that could be required in some situations are `future.seed`, `future.scheduling`, `future.globals.maxSize`, and `future.resolve.recursive`. So, it's always better to leave the `future` setup on the users to customize it for their machines. None-experienced users can rely on `?future::plan`.

hsbadr:
@AlejandroCatalina For this PR, would it be cleaner to move `refmodel$fit_elapsed_time` to `refmodel$fit$elapsed_time`? i.e., adding an `elapsed_time` slot to the fit object.

hsbadr:
3206c839242a450b2885b4b66285ddf73871e303 makes the squashed changes minimal (just add `elapsed_time` slot to the `refmodel$fit` and use it).

AlejandroCatalina (changes requested):

R/refmodel.R
```diff
@@ -459,12 +463,6 @@ init_refmodel <- function(object, data, formula, family, ref_predfun = NULL,
   target <- .get_standard_y(y, weights, family)
   y <- target$y
 
-  if (inherits(object, "brmsfit")) {
-    object$elapsed_time <- rstan::get_elapsed_time(object$fit)
```
AlejandroCatalina: We can keep this here until brms returns `object$elapsed_time`.

hsbadr: Ok. We can add it in the `get_refmodel.brmsfit`.

Do you wanna keep this `brmsfit` check: https://github.com/hsbadr/projpred/blob/39b5321e74f571b8e458316652d56588632de964/R/cv_varsel.R#L803

hsbadr: https://github.com/paul-buerkner/brms/pull/1136/commits/2b83b4ee7b3aa6d267c1042d915f0b13303382db should do it.

AlejandroCatalina: We can wait to merge this until that merge goes through on brms.

hsbadr: Sounds good.

hsbadr:
@AlejandroCatalina You may want to unify the user interface (or at least combine all `varsel_cv` and `cv_varsel` into a single function). I think we could also merge them into `varsel` with the different variants controlled by `cv_method` argument: e.g. if `is.null(cv_method)` run `varsel` without cross validation, and if a CV method is specified use it.

AlejandroCatalina:
That's something on our mind of course, but at this point this interface is not yet ready to fully replace the current one, although we are working to make that so in the near future. The idea is that calling `varsel_cv` can invoke any of the cross validation possibilities for a time, but in the end we want `varsel_cv` to represent full CV while `approx_cv` represents approximate CV. We need to keep thinking about the interface from the user's perspective.

hsbadr:
> That's something on our mind of course, but at this point this interface is not yet ready to fully replace the current one, although we are working to make that so in the near future. The idea is that calling `varsel_cv` can invoke any of the cross validation possibilities for a time, but in the end we want `varsel_cv` to represent full CV while `approx_cv` represents approximate CV. We need to keep thinking about the interface from the user's perspective.

I see. I'd take care of all of this via a wrapper function `varsel` that calls the appropriate method via an argument like:
```r
cv_method = c("none", "loo", "loo_approx", "kfold", "kfold_approx")
```
The current `varsel` function can be renamed with `cv_method = "none"` as a default.

AlejandroCatalina (approved):

AlejandroCatalina:
Yeah, that's a good suggestion. That would make `varsel_cv` able to replicate all current behaviour while keeping the independent `approx_{cv}` functions that are nicer from the workflow perspective.

Regarding the current PR, I'll go ahead and merge it when the tests pass.

AlejandroCatalina:
I merged this by hand in ffc0e2c

PR #130, opened 2021-05-20
Fix `as.matrix.projection()` for 1 (clustered) draw

fweber144:
This fixes the following issue:
```r
library(projpred)
library(rstanarm)
options(mc.cores = parallel::detectCores(logical = FALSE))
data("df_gaussian", package = "projpred")
df_gaussian <- df_gaussian[1:40, ]
mydat <- cbind("y" = df_gaussian$y, as.data.frame(df_gaussian$x))
mydat$mygroup <- gl(n = 8, k = floor(nrow(mydat) / 8),
                    labels = paste0("gr", seq_len(8)))
set.seed(457211)
mydat$noise <- rnorm(nrow(mydat))
mygroup_icpts_truth <- rnorm(nlevels(mydat$mygroup), sd = 6)
mygroup_V1_truth <- rnorm(nlevels(mydat$mygroup), sd = 6)
mydat$y <- mydat$y + mygroup_icpts_truth[as.numeric(mydat$mygroup)]
mydat$y <- mydat$y + mygroup_V1_truth[as.numeric(mydat$mygroup)] * mydat$V1

myfit <- stan_glmer(y ~ V1 + V2 + V3 + V4 + V5 + noise + (1 + V1 | mygroup),
                    data = mydat,
                    seed = 1140350788)

myproj <- project(myfit,
                  solution_terms = c("(1 | mygroup)"),
                  nclusters = 1)
myprjmat <- as.matrix(myproj)
str(myprjmat)
## Previously (on branch `develop`), this gave:
# List of 2
# $ :'data.frame':	8 obs. of  1 variable:
#   ..$ (Intercept): num [1:8] 2.25 3.3 1.23 -6.04 8.19 ...
# $ : num 9.03
# - attr(*, "dim")= int [1:2] 1 2
# - attr(*, "dimnames")=List of 2
# ..$ : NULL
# ..$ : chr [1:2] "b_mygroup" "sigma"
##
## With the fix from this PR, this gives:
# num [1, 1:11] 2.2269 5.8988 0.0264 1.0743 -0.9944 ...
# - attr(*, "dimnames")=List of 2
# ..$ : NULL
# ..$ : chr [1:11] "b_Intercept" "sd_mygroup__Intercept" "r_mygroup[gr1,Intercept]" "r_mygroup[gr2,Intercept]" ...
##

```
[...]

Issue #131, opened 2021-05-21
Error for `as.matrix()` applied to GAMM projection


fweber144:
Applying `as.matrix()` to a GAMM projection does not work:
```r
# Source: Example section of `?rstanarm::stan_gamm4`.
library(projpred)
library(rstanarm)
options(mc.cores = parallel::detectCores(logical = FALSE))
dat <- mgcv::gamSim(1, n = 400, scale = 2) ## simulate 4 term additive truth
## Now add 20 level random effect`fac'...
dat$fac <- fac <- as.factor(sample(1:20, 400, replace = TRUE))
dat$y <- dat$y + model.matrix(~ fac - 1) %*% rnorm(20) * .5
br <- stan_gamm4(y ~ s(x0) + x1 + s(x2), data = dat, random = ~ (1 | fac),
                 chains = 1,
                 iter = 500, # for example speed
                 seed = 1140350788)

myproj <- project(br,
                  solution_terms = c("x1", "s(x2)", "(1 | fac)"),
                  nclusters = 2)
myprjmat <- as.matrix(myproj)
```
On branch `develop` (for example), that last line raises the following error:
```
Error in names(population_effects) <- replace_intercept_name(names(population_effects)) :
  attempt to set an attribute on NULL
```
The issue is probably that `as.matrix.lm()` (which is used for submodels of class `"gamm4"`) uses `coef()` which in turn is `NULL` for objects of class `"gamm4"`:
```r
coef(myproj$sub_fit[[1]])
## --> NULL
```
So for submodels of class `"gamm4"`, the `as.matrix.lm()` method probably can't be used.

AlejandroCatalina:
Yes, I see. Why would `gamm4` objects not implement `coef`? It seems it's a fairly generic method that could be used in this case as well, but anyway. I'll look at the PR.

fweber144:
Yes, a corresponding `coef()` method would be helpful. But the documentation of `gamm4::gamm4()` says:

> Returns a list with two items:
>   * `gam`: an  object  of  class `gam`. [...]
>   * `mer`: the fitted model object returned by `lmer` or `glmer`. [...]

Perhaps you can exploit that?

AlejandroCatalina:
Yes. I think we can possibly merge both coefficients matrices together. I'll play with it locally and see what comes out.

PR #132, opened 2021-05-22
Use brms's naming convention also for submodels of class `"subfit"`

fweber144:
Since commit e420740961529c2f13fc26675b0e6ba30cf94a32 ff., the `as.matrix.<class_name>()` methods use brms's naming convention for the parameters. However, the `as.matrix.subfit()` method was somehow forgotten at that time, as illustrated by the following reprex:
```r
# Source: <https://github.com/stan-dev/projpred/blob/master/README.md>
library(projpred)
library(rstanarm)
myfit <- stan_glm(y ~ V1 + V2 + V3 + V4 + V5,
                  family = gaussian(),
                  data = mydat,
                  prior = hs(df = 1, global_scale = 0.01),
                  seed = 1140350788)

myprj <- project(myfit,
                 solution_terms = c("V1"),
                 nclusters = 2)
# Only to show that the submodels are of class "subfit":
stopifnot(identical(
  class(myprj$sub_fit[[1]]),
  c("subfit")
))
myprjmat <- as.matrix(myprj)
colnames(myprjmat)
```
Previously (i.e. on branch `develop`, for example), that last line gave:
```
[1] "Intercept" "V1"        "sigma"
```
This PR fixes that issue (i.e. it uses brms's naming convention also for submodels of class `"subfit"`), so that the reprex's last line gives:
```
[1] "b_Intercept" "b_V1"        "sigma"
```

Note that commit 6e56e6c is basically taken from PR #130. It is included here to have a clear history.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -24,7 +24,7 @@
 * Fixed bugs for argument `nterms` of `proj_linpred()` and `proj_predict()`. (GitHub: #110)
 * Fixed an inconsistency for some intercept-only submodels. (GitHub: #119)
 * Minor documentation fixes.
-* For submodels of class `"subfit"`, make the column names of `as.matrix.projection()`'s output matrix consistent with other classes of submodels. (GitHub: #<INSERT_PR_NUMBER>)
+* For submodels of class `"subfit"`, make the column names of `as.matrix.projection()`'s output matrix consistent with other classes of submodels. (GitHub: #132)
 
 ## projpred 2.0.5
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -24,6 +24,7 @@
 * Fixed bugs for argument `nterms` of `proj_linpred()` and `proj_predict()`. (GitHub: #110)
 * Fixed an inconsistency for some intercept-only submodels. (GitHub: #119)
 * Minor documentation fixes.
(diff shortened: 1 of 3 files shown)


AlejandroCatalina:
This is good thanks! Can we add tests for the names of the resulting coefficients matrices after `project` for all model types currently supported by this naming convention?

fweber144:
Yes, I'll combine those tests with those from PR #130.

fweber144:
I've added the requested tests (combined with those from PR #130). Please take a look if everything is OK for you this way.

fweber144:
As soon as issue #131 is fixed, we can add a GAMM to `test_as_matrix.R`.

AlejandroCatalina:
Why not use `test_that`?

fweber144:
Well, I do.

fweber144:
I just refactored `test_as_matrix.R` in commit 370a3df003cedb6d8b3c80b2b61b638b14d13ad8 so I could extend it more easily (here and in the future as well).

AlejandroCatalina:
Yeah I just missed it at first glance. Thanks!

Best,
Alejandro

fweber144:
No problem. Just FYI: The `expect_warning()` calls can't be wrapped in `test_that()` calls since otherwise, the objects created within them wouldn't be accessible later.

PR #134, opened 2021-05-22
Handle `size_sub` differently and fix `seed_sub`

fweber144:
This fixes the following issue:
```r
library(projpred)
library(rstanarm)
myfit <- stan_glm(y ~ V1 + V2 + V3 + V4 + V5,
                  family = gaussian(),
                  data = mydat,
                  prior = hs(df = 1, global_scale = 0.01),
                  seed = 1140350788)

mypred <- proj_predict(myfit,
                       solution_terms = c("V3", "V2"),
                       newdata = mydat[1:3, , drop = FALSE],
                       nclusters = 2)
str(mypred)
```
Previously (i.e. on branch `develop`, for example), that last line gave:
```
num [1:1000, 1:3] 2.834 -1.349 -0.357 0.392 3.44 ...
```
This PR fixes that issue (i.e. it cuts off `size_sub` at its maximum possible value), so that the reprex's last line gives:
```
num [1:2, 1:3] 0.0801 1.9877 0.6548 2.1995 -0.9375 ...
```
**EDIT:** Please read the comments below before merging.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -24,6 +24,7 @@
 * Fixed bugs for argument `nterms` of `proj_linpred()` and `proj_predict()`. (GitHub: #110)
 * Fixed an inconsistency for some intercept-only submodels. (GitHub: #119)
 * Minor documentation fixes.
+* Argument `size_sub` of `proj_predict()` is now cut off at its maximum possible value. (GitHub: #134)
 
 ## projpred 2.0.5
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -24,7 +24,8 @@
 * Fixed bugs for argument `nterms` of `proj_linpred()` and `proj_predict()`. (GitHub: #110)
 * Fixed an inconsistency for some intercept-only submodels. (GitHub: #119)
 * Minor documentation fixes.
-* Argument `size_sub` of `proj_predict()` is now cut off at its maximum possible value. (GitHub: #134)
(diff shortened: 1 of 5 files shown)


fweber144:
If the previous behavior was chosen on purpose (because of `replace = TRUE` in the corresponding `sample()` call), just ignore this PR.

fweber144:
The more I think about it, the more I think it would make sense to differentiate between thinned and clustered posterior draws. For the former, I think all projected draws should be taken (without offering the option to sample from them). For the latter, I think the previous behavior (i.e. before this PR) with `replace = TRUE` and not cutting off `size_sub` at `NCOL(mu)` was appropriate.

As a next step, it could make sense to offer this differentiation at other places, too: e.g. in `as.matrix.projection()` (which currently warns in case of clustered posterior draws) or in `proj_linpred(..., integrated = FALSE)`.

fweber144:
The last 4 commits I just pushed should handle the differentiation between thinned and clustered posterior draws described above. And they (more precisely: commit 706f18e) also fix an issue for argument `seed_sub`.

fweber144:
Sorry, its seems like I messed up `methods.R` by introducing some roxygen2 comments which are not supposed to be there. And since this PR has evolved a lot, I'll close it and open up a new one. Sorry for the confusion.

PR #139, opened 2021-05-26
Fix error for L1 search applied to an empty (i.e. intercept-only) reference model

fweber144:
This fixes the following issue:
```r
library(projpred)
library(rstanarm)
myfit <- stan_glm(y ~ 1,
                  data = mydat,
                  seed = 1140350788)
my_vs <- varsel(myfit, nclusters = 3, nclusters_pred = 5, method = "L1")
```
That last line threw the uninformative error
```
Error in seq.default(log(lambda_min), log(lambda_max), len = nlam) :
  'from' must be a finite number
In addition: Warning messages:
  1: In max(lambda_max_cand[is.finite(lambda_max_cand)]) :
  no non-missing arguments to max; returning -Inf
2: In log(lambda_min) : NaNs produced
3: In log(lambda_max) :
```
Now it throws the more appropriate error
```
Error in search_L1(p_sel, refmodel, family, intercept, nterms_max - intercept,  : 
  L1 search cannot be used for an empty (i.e. intercept-only) reference model.
```
The problem for this special case is that in lines https://github.com/stan-dev/projpred/blob/fb16ff3304e38c94fed3c4ef3924022d936c745a/R/glmfun.R#L110-L111 we have `length(lambda_max_cand) == 0`.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -26,6 +26,7 @@
 * Minor documentation fixes.
 * Fix a bug for `as.matrix.projection()` in case of 1 (clustered) draw after projection. (GitHub: #130)
 * For submodels of class `"subfit"`, make the column names of `as.matrix.projection()`'s output matrix consistent with other classes of submodels. (GitHub: #132)
+* Throw an appropriate error message when trying to apply an L1 search to an empty (i.e. intercept-only) reference model. (GitHub: #139)
 
 ## projpred 2.0.5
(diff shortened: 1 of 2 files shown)


AlejandroCatalina:
Can you fix the conflicts on NEWS.md?

fweber144:
Done

Issue #159, opened 2021-06-21
Projection onto intercept-only model gives different `solution_terms`


fweber144:
Consider the following reprex (tested on branch `develop`):
```r
library(rstanarm)
options(mc.cores = parallel::detectCores(logical = FALSE))
data("df_gaussian", package = "projpred")
mydat <- cbind("y" = df_gaussian$y, as.data.frame(df_gaussian$x))
myfit <- stan_glm(y ~ V1 + V2 + V3 + V4 + V5,
                  family = gaussian(),
                  data = mydat,
                  prior = hs(df = 1, global_scale = 0.01),
                  seed = 1140350788)
library(projpred)
myvsel <- varsel(myfit, method = "forward", nterms_max = 0,
                 nclusters = 3, nclusters_pred = 5)
```
Now
```r
solution_terms(myvsel)
```
gives
```
character(0)
```
whereas
```r
myprj <- project(myfit, solution_terms = character(), nclusters = 5)
myprj$solution_terms
```
gives
```
[1] "1"
```
I don't know if this inconsistency has any consequences for downstream code, but already from a user perspective, it would be good to have a consistent output. As another step, one could also think about always including `"1"` since
```r
myprj_V1 <- project(myfit, solution_terms = c("V1"), nclusters = 5)
myprj_V1$solution_terms
```
for example gives
```
[1] "V1"
```
so it doesn't include the intercept.

PR #165, opened 2021-07-06
Fix the fix for issue #159

fweber144:
This PR ensures that after the fix for #159 (commit a422b9b7d20ed521d1293354a8dff99bb95d361d), `solution_terms` is still always of type `character`, even for an empty submodel. (The fix for #159 was using `c()` but that is equivalent to `NULL`.) It also fixes an issue with `filter_nterms` caused by the fix for #159.

### DIFF
--- a/R/methods.R
+++ b/R/methods.R
@@ -104,7 +104,11 @@ proj_helper <- function(object, newdata,
         object <- list(object)
       }
       projs <- Filter(function(x) {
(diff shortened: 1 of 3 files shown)


fweber144:
Just realizing that the issue with `filter_nterms` is caused by a deeper issue with `count_terms_chosen()` caused by the fix for #159. I'll close this, try to fix `count_terms_chosen()`, and then create a new PR.

PR #201, opened 2021-08-17
Enhance/fix `fit_glmer_callback()`

fweber144:
This PR enhances (and possibly fixes) `fit_glmer_callback()` in the following ways:

* It handles the error `"pwrssUpdate did not converge in (maxit) iterations"` (or at least it tries to). At least in the examples I have run, the advice given [here](https://github.com/lme4/lme4/issues/573#issuecomment-802267549) worked.
* It avoids the `preprocess_data()` call since:
  - scaling predictors changes their regression coefficients;
  - `preprocess_data()` is not adopted to categorical predictors;
  - `preprocess_data()`'s implementation for interactions is probably incorrect (the interaction columns in the model matrix must not be scaled; they need to be re-calculated from the scaled inputs, see [Gelman (2008, section 3.1)](https://doi.org/10.1002/sim.3107)).
  
  At least in the examples I have run, avoiding `preprocess_data()` is possible due to the advice given [here](https://github.com/lme4/lme4/issues/579#issuecomment-662789003).

Note that there is another `preprocess_data()` call in `fit_gamm_callback()` which probably should and perhaps could also be avoided. If not, I think it needs at least a warning when `preprocess_data()` is applied (so the users knows of the modified regression coefficients).

### DIFF
--- a/R/divergence_minimizers.R
+++ b/R/divergence_minimizers.R
@@ -157,6 +157,12 @@ fit_glmer_callback <- function(formula, data, family, weights,
         formula, data = data, family = family, weights = weights,
         control = control, nAGQ = 20L, ...
       ))
+    } else if (grepl("pwrssUpdate did not converge in \\(maxit\\) iterations",
+                     as.character(e))) {
+      return(fit_glmer_callback(
+        formula, data = data, family = family, weights = weights,
+        control = control_callback(family, tolPwrss = 1e-6), ...
+      ))
     } else {
       stop(e)
     }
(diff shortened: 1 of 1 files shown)


AlejandroCatalina:
Would the same apply for gamms? I would say so?

fweber144:
Yes, the same principles apply for GAMMs, I would say. But I don't know if my fixes for GLMMs can be easily transferred to GAMMs.

fweber144:
(For example, I don't know if `gamm4::gamm4()` accepts argument `nAGQ` and if yes, if that also solves the `"not positive definite"` error for which `preprocess_data()` has been added in `fit_gamm_callback()`.)

AlejandroCatalina:
Yes. Okay I think we can merge this one and work on the other one later on.

Best,
Alex
________________________________

fweber144:
Yes, I'll create an issue for the GAMM part.

PR #170, opened 2021-07-13
Fix tests

fweber144:
Following the fixes of issues #158 and #159, this PR adopts the tests accordingly. All tests should be passing now.

**EDIT:** I also had to adopt `test_varsel.R` to the fix for issue #167.

PR #209, opened 2021-09-23
Bug fix in `proj_linpred()`

fweber144:
This fixes a bug in `proj_linpred()`: For `integrated = TRUE` and `is.null(ynew)`, a `if (!is.null(lpd_out))` condition had to be added before averaging `lpd_out`.

### DIFF
--- a/R/methods.R
+++ b/R/methods.R
@@ -232,7 +232,11 @@ proj_linpred_aux <- function(proj, mu, weights, ...) {
   if (dot_args$integrated) {
     ## average over the posterior draws
     pred_out <- pred_out %*% proj$weights
-    lpd_out <- as.matrix(apply(lpd_out, 1, log_weighted_mean_exp, proj$weights))
+    if (!is.null(lpd_out)) {
+      lpd_out <- as.matrix(
+        apply(lpd_out, 1, log_weighted_mean_exp, proj$weights)
+      )
+    }
   }
   return(nlist(pred = t(pred_out),
                lpd = if(is.null(lpd_out)) lpd_out else t(lpd_out)))


fweber144:
Thanks for merging this. To give a rough idea of what I would like to come up with next:

1. As soon as the Travis CI issue mentioned [here](https://github.com/stan-dev/projpred/pull/170#issuecomment-878849775) is resolved, I'll create a minor PR fixing just a small issue with `print.vsel()`. I would wait for the Travis CI issue to be resolved so that we get an automatic coverage report for branch `develop`.
2. A PR with completely refactored tests. With the Travis CI issue being resolved, we would get an automatic coverage report for the refactored tests, too. (And the aim is to have a coverage comparable to the state before, but at a smaller computing time.)
3. Several PRs with some fixes and minor modifications.
4. A PR extending the refactored tests to cover even more scenarios (at an increased computational cost, but I think that's ok).
5. A PR completely rewriting the divergence minimizers (replacing them by a single one) so it's easier to introduce the parallelization of the projection and also the new divergence minimizer for the augmented-data projection later. The rewritten divergence minimizer should also have less issues with the automatic handling of errors from the underlying external packages, at least for submodels which are GLMMs.
6. Some minor PRs.
7. A PR completely rewriting the documentation (and using Markdown syntax there).
8. One minor PR.
9. A PR completely rewriting the vignettes (replacing them by a single new one) so that this new vignette can be built faster (and includes the code output), which is important for the next CRAN release. I have also added the `posterior` package as another post-processing option in the new vignette.
10. A PR updating the NEWS file and increasing the version number (and date).

Then, I think, we should be prepared for a new CRAN release. Perhaps I can already include the PR introducing the parallelization of the projection in that CRAN release. This depends on how much work it is to update the tests after the parallelization. The rewritten vignette could then also take advantage of the parallelized projection, at least with 2 cores which should also be available on CRAN.

Issue #70, opened 2021-01-20
Discrete families with finite support (and possibly other families)


fweber144:
It would be great to have more outcome families supported by projpred. For discrete outcome families with finite support (e.g. the cumulative family in ordinal regression), the projection can be performed via a pseudo-dataset approach:

For these families, equation (12) from 

Piironen J, Paasiniemi M, Vehtari A (2020). Projective inference in high-dimensional problems: Prediction and feature selection. *Electronic Journal of Statistics*, 14(1), 2155–2197. https://doi.org/10.1214/20-EJS1711

simplifies to

![formula1](https://user-images.githubusercontent.com/55132727/105222399-b8c47880-5b5a-11eb-894a-05c8fa6bf28d.png)

with supp(\tilde{y}) denoting the (discrete and finite) support of the outcome distribution (i.e. the set of possible values for the outcome y).

This simplification corresponds to a weighted maximum-likelihood problem with weights

![formula2](https://user-images.githubusercontent.com/55132727/105223408-07bedd80-5b5c-11eb-92d3-0a4a6f8ebb4b.png)

and a pseudo (or artificial) dataset constructed as follows:

1. Take the original dataset. Denote by n the number of rows in this original dataset.
2. Repeat each row another (|supp(\tilde{y})| - 1) times. The resulting dataset then has n * |supp(\tilde{y})| rows.
3. Replace the outcome variable (column "response", say) by all possible values for y (i.e. each value from supp(\tilde{y})) such that each row i \in {1, ..., n} from the original dataset occurs together with each possible value for y (in column "response") exactly once.

The weight for the pseudo-dataset row coming from row i \in {1, ..., n} in the original dataset and having outcome value y is then a_i^*(y) as defined above.

Thus, it should be possible to use existing routines for solving this weighted maximum-likelihood problem, like `MASS::polr()` for the cumulative family (in case of no group-level or smoothing terms). PSIS-LOO-CV also naturally fits into the simplification above (simply introducing the importance weights when averaging over the posterior draws).

Of course, the approach is primarily useful for outcome families (discrete and finite support) which do not constitute an exponential family (since exponential families may be treated the usual way, see Piironen et al., 2020).

**EDITS:**

- Minor wording improvements in the text above.
- The approach might also be used as an approximation for discrete families with infinite support if support areas with near-zero probability are ignored.
- In principle, a similar pseudo-data approach could also be used for any outcome family (including continuous ones and also discrete ones with infinite support) with one of the following two approaches:
[...]

PR #217, opened 2021-10-01
Refactored tests

fweber144:
This PR refactors the unit tests in order to:

* speed them up,
* consistently add tests for GLMMs,
* generalize them in preparation for the augmented-data projection discussed in #70.

The speed improvement is achieved by creating a single setup file (`tests/testthat/setup.R`) which is automatically sourced by **testthat** at the beginning of the tests (and only once). All refactored test files `tests/testthat/test_*.R` use the objects created in this `setup.R` file (as far as they need them).

On my machine, the tests take about 131s when launched via the RStudio keyboard shortcut, i.e., using `devtools::test()`. On branch `develop` (commit 22c61cf610458caf1f993e2043b9b05c6a34adad), the tests take about 437s when launched in the same way. Coverage (computed locally; script below) is comparable: I get a coverage of 79.82% for the refactored tests and a coverage of 79.29% for branch `develop`. The refactored tests will be extended in the future and should then achieve a higher coverage, but I am deferring that to a later PR because some bugs have to be fixed first and because I wanted to demonstrate that a comparable coverage can be achieved at a much smaller computing time.

Remarks:

1. Files I have left untouched (apart from cleaning up afterwards):
    + `test_cvindices.R` (runs very fast; doesn't need to be generalized to the augmented-data projection)
    + `test_formula.R` (runs very fast)
    + `test_glm_elnet.R` (takes a bit of time, but not too long; doesn't need to be generalized to the augmented-data projection)
    + `test_glm_ridge.R` (runs very fast; doesn't need to be generalized to the augmented-data projection)
1. In `test_datafit.R`, I have left the comparison with package **glmnet** largely untouched because it relies on hard-coded expected values (and because it runs fast and doesn't need to be generalized to the augmented-data projection).
1. There are some to-dos (indicated by `TODO` comments in the test files).
1. The comment
    ```r
    # TODO: Fix this: When using all `fam_nms`, use the following order because [...]
    ```
    will be fixed by a PR which I have already prepared in my fork.
1. The comment
    ```r
    # TODO: This might in fact be undesired
    ```
    is related to paul-buerkner/brms#1230. I'll create an issue for that.
1. Commit
    ```
    Ignore vignettes in `R CMD build` and `R CMD check`. Only check a single architecture in `R CMD check`.
    ```
    is not really related to the refactored tests (I could have created a separate PR for that, but that seemed a bit over the top).

Script for calculating the coverage locally:
```r
Sys.setenv("NOT_CRAN" = "true")
library(covr)
print(system.time(pkgcov <- package_coverage()))
print(percent_coverage(pkgcov))
covreprt_path <- report(pkgcov)
```

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -6,3 +6,4 @@
 ^vignettes/quickstart_cache$
 ^man-roxygen$
 ^tests/testthat/bfits$
+^tests/testthat/_snaps$
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ vignettes/quickstart_cache
 *cache/
 *files/
 tests/testthat/bfits
+tests/testthat/_snaps
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -48,7 +48,8 @@ Suggests:
     rmarkdown,
     glmnet,
     bayesplot (>= 1.5.0),
-    optimx
+    optimx,
+    digest
 VignetteBuilder: knitr
 URL: https://mc-stan.org/projpred/, https://discourse.mc-stan.org
 BugReports: https://github.com/stan-dev/projpred/issues/
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -48,8 +48,7 @@ Suggests:
     rmarkdown,
     glmnet,
     bayesplot (>= 1.5.0),
-    optimx,
-    digest
+    optimx
 VignetteBuilder: knitr
 URL: https://mc-stan.org/projpred/, https://discourse.mc-stan.org
 BugReports: https://github.com/stan-dev/projpred/issues/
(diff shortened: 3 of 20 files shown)


fweber144:
> 5. The comment
>     ```r
>     # TODO: This might in fact be undesired
>     ```
>     is related to [`extract_model_data()`: ignored arguments paul-buerkner/brms#1230](https://github.com/paul-buerkner/brms/issues/1230). I'll create an issue for that.
>

This may now be found in #218.

fweber144:
I just realized that I somehow confounded "adopt" and "adapt" in commit messages of this PR (and also in previous PRs, I think). So what reads "adopt" should actually read "adapt". Sorry for the confusion.

PR #223, opened 2021-10-07
Miscellaneous modifications and fixes

fweber144:
This PR performs some minor enhancements, including some bug fixes. See the commit messages for details.

### DIFF
--- a/R/methods.R
+++ b/R/methods.R
@@ -732,7 +732,7 @@ suggest_size <- function(object, ...) {
 #' }
 #'
 #' @export
-suggest_size.vsel <- function(object, stat = "elpd", alpha = 0.32, pct = 0.0,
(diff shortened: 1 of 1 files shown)


fweber144:
I forgot to mention the following:

* `as.matrix.projection()` doesn't seem to be possible for `datafit`s (where `project()` has been applied to an object of class `vsel`). I don't know if this is on purpose. For this PR, I was assuming yes, so we now throw an appropriate error message.
* `project()` doesn't seem to be possible for `datafit`s directly. I don't know if this is on purpose. For this PR, I was assuming yes, so we now throw an appropriate error message.
* For this PR, I was assuming that for `datafit`s, the default of argument `cv_search` in `varsel()` and `cv_varsel()` should be `FALSE` because `project()` behaves that way and because the outdated `is.null(cv_search)` condition in `parse_args_varsel()` hinted at that behavior.

PR #226, opened 2021-10-07
Docs: Activate Markdown syntax

fweber144:
This activates Markdown syntax in the documentation. Here, no content is changed (apart from the minor formatting changes mentioned in the commit messages). The content changes will be the subject of another PR.


fweber144:
Maybe I should have added a resource where the Markdown syntax in **roxygen2** is explained. You may find such information in [this vignette](https://cran.r-project.org/web/packages/roxygen2/vignettes/rd-formatting.html) (which should be the same as [this one](https://roxygen2.r-lib.org/articles/markdown.html), but currently, the latter seems to refer to an older version).

PR #233, opened 2021-10-10
Major enhancements in the docs

fweber144:
This PR provides a complete overhaul of the documentation. It not only makes the docs use Markdown syntax (consistently), but also introduces important enhancements regarding the content. It would be good if from now on, all future documentation in **projpred** used Markdown syntax. That way, it would stay consistent and searching for specific words/expressions would be made easier.

I still have some questions:

1. In `summary.vsel()` and `suggest_size.vsel()`,  shouldn't the term "credible" (intervals/bounds) be replaced by "confidence" (intervals/bounds)? If yes, then I think probability statements such as https://github.com/stan-dev/projpred/blob/7e67eed52cd307954f47cea939a6a595b07e1bfb/R/methods.R#L701-L704 are incorrect since the interpretation has to be frequentist, i.e. based on infinite repetitions. I can change the documentation correspondingly if you want. **EDIT:** Also see the [comment below](https://github.com/stan-dev/projpred/pull/233#issuecomment-939494544).
1. In `summary.vsel()`, argument `type`: The lower and upper bounds are for a *central* interval, right? Then I would change the documentation to be more precise in that regard.
1. The documentation for the `mesquite` dataset refers to `Gelman and Hill (2007)`, but a corresponding reference is missing. What is the correct reference for this?
1. I found this: https://github.com/stan-dev/projpred/blob/7e67eed52cd307954f47cea939a6a595b07e1bfb/R/families.R#L16-L18 How do you want me to handle this?
1. If the Student-*t* family should indeed be supported (see the previous enumeration point), then doesn't the Student-*t* family also need to be taken into account in the case https://github.com/stan-dev/projpred/blob/7e67eed52cd307954f47cea939a6a595b07e1bfb/R/projfun.R#L106 of `.init_submodel()`?

Tell me if these questions can't be answered easily and you prefer separate issues for them.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ warnings_are_errors: false
 
 latex: true
 r_build_args: '--no-build-vignettes'
-r_check_args: '--ignore-vignettes --no-examples'
+r_check_args: '--ignore-vignettes'
 
 r_github_packages:
   - jimhester/covr
@@ -30,7 +30,7 @@ before_install:
 
 script:
   - travis_wait 120 R CMD build . --no-build-vignettes
-  - travis_wait 120 R CMD check projpred*tar.gz --ignore-vignettes --no-examples
+  - travis_wait 120 R CMD check projpred*tar.gz --ignore-vignettes
 
 after_success:
 - travis_wait 120 Rscript -e 'covr::codecov()'


AlejandroCatalina:
Yes, confidence is the right term here.

Best,
Alex
________________________________

AlejandroCatalina:
I don’t think so at this point but I have not tested it.

Best,
Alex
________________________________

fweber144:
Ok, so I'll remove the Student-*t* family in a future PR.

fweber144:
> Ok, so I'll remove the Student-_t_ family in a future PR.

I decided to throw just a warning, not an error, to avoid a breaking change. See PR #252.

PR #235, opened 2021-10-11
Parallelization of the projection

fweber144:
This introduces the parallelization of the projection as an optional feature. By default, that parallelization is turned off. Corresponding documentation may be found in the general package documentation (available at `` ?`projpred-package` ``).

Note that parallelizing the projection doesn't work well on Windows because in my experience, the parallelization overhead is larger there, causing a parallel run to take longer than a sequential run.

Also note that the parallelization works well for GLMs, but for GLMMs, GAMs, and GAMMs, the fitted model objects are quite big, which—when running in parallel—may lead to an excessive memory usage which in turn may crash the R session. I have attempted to circumvent this issue, but all my attempts have been in vain. So the only solution would be to reduce the return values of the corresponding `fit_<...>_callback()` functions so they only consist of the corresponding parameter estimates (essentially a single row of the matrix returned by `as.matrix.projection()`). The downside is that reducing the return values makes the use of predefined functions like `lme4:::predict.merMod()` impossible. So we would have to write custom `predict()` methods which are able to calculate the linear predictor based on the reduced return value and a (possibly new) dataset.

### DIFF
--- a/tests/testthat/setup.R
+++ b/tests/testthat/setup.R
@@ -60,10 +60,8 @@ if (run_prll) {
             "could be detected.")
     run_prll <- FALSE
   }
-  # Just to be sure to not run on more than 2 cores on CRAN (shouldn't be
-  # necessary, though, because by default, we do not run the parallel tests on
-  # CRAN; see the definition of `run_prll` above):
-  if (!identical(Sys.getenv("NOT_CRAN"), "true")) {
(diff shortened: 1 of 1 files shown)


AlejandroCatalina:
Yeah, I’ve had this issue too when implementing a parallel approach. We really need to work on downsizing the output values. For now this should so.

Issue #137, opened 2021-05-25
Poisson example in GLMM vignette


fweber144:
Even after #133, the Poisson example in the GLMM vignette stays strange: For the hierarchical SD of `(1 | obs)`, the projection gives exactly 1 across all projected draws. When looking closer at the submodel from the first projected draw, I got this output:
```
Generalized linear mixed model fit by maximum likelihood (Laplace Approximation) ['glmerMod']
 Family: poisson  ( log )
Formula: .phen_pois.1 ~ cofactor + (1 | obs)
   Data: data
Weights: weights
     AIC      BIC   logLik deviance df.resid 
     Inf      Inf     -Inf      Inf      197 
Random effects:
 Groups Name        Std.Dev.
 obs    (Intercept) 1       
Number of obs: 200, groups:  obs, 200
Fixed Effects:
(Intercept)     cofactor  
    -1.9833       0.2467  
optimizer (Nelder_Mead) convergence code: 0 (OK) ; 14000 optimizer warnings; 1 lme4 warnings
```
The optimizer and lme4 warnings there at the end might already indicate an issue. Perhaps it's the try to model overdispersion by including `(1 | obs)` (where `obs` has a one level per observation) which causes problems here. (Note that in the [currently published vignette](https://mc-stan.org/projpred/articles/quickstart_glmm.html#poisson-example-1), it *seems* like the hierarchical SD is not exactly 1, but that's only due to the former use of the empirical SD of the random effects instead of their hierarchical SD, see #74.)

Anyway, since the vignette already takes very long to run through, perhaps simply omit this Poisson example? Or completely re-organize the whole GLMM vignette? In my opinion, a single example would be enough. And since the Bernoulli family is probably the most common one for *generalized* linear (mixed) models, I would choose a Bernoulli example. The `VerbAgg` example from the current vignette is nice, but it has a lot of observations (even when restricting it to 50 individuals) and the varsel (as well as the CV varsel) plot looks strange. Perhaps consider the `MASS::bacteria` dataset instead, e.g. with the formula `y ~ week * trt + (1 | ID)`?

AlejandroCatalina:
That’s reasonable. We just wanted to include enough variation but I agree one example would be enough. I don’t have time right now to remake the vignette so it’ll have to stay for a while at least. If you want to give it a shot I can help with anything.

Best,
Alejandro

fweber144:
[...]

Issue #208, opened 2021-09-22
Question concerning statement in GLMM vignette


fweber144:
The GLMM vignette says:

> For non-Gaussian models the source of the inaccuracy may come from the fact that we are running an approximate projection. In this case, increasing the number of draws for the prediction `ndraws_pred` should also help.

Does "approximate" here refer to the fact that the defaults of `ndraws` and `nclusters` (as well as `ndraws_pred` and `nclusters_pred`) do not imply a draw-by-draw projection, meaning that not all posterior draws are projected or that they are clustered before they are projected?

fweber144:
Ah, now that I read [this preprint](https://arxiv.org/abs/2109.04702v1), could it be that you were talking of what is mentioned in section 3.2 of that preprint, i.e., the fact that the PIRLS algorithm might have convergence issues for non-Gaussian families (also shown/discussed in sections 5.4 and 5.5)?

fweber144:
Closing due to inactivity. (In #237, I made the adjustments in the vignette by guessing what was probably meant.)

PR #237, opened 2021-10-12
Vignette(s)

fweber144:
This replaces the two previous vignettes (GLM and GLMM) by a single new vignette. In that new vignette, I tried to merge the contents from the two previous vignettes (as far as possible). The new vignette used to have a GLMM example (the `datasets::CO2` dataset, see the commit messages), but due to the problems in parallelizing the projection for GLMMs (see #235) as well as the fact that the reference model for the `CO2` dataset performed badly in its PPC, I switched back to the GLM example (the `projpred::df_gaussian` data) from the former GLM vignette and then introduced the parallelization (on non-Windows systems). Notes/to-dos:

* This requires an update of the <https://mc-stan.org/projpred/> website. (In the general package documentation available at `` ?`projpred-package` `` as well as in an upcoming update of the `README` file, I introduced the link to the new vignette on the mc-stan.org/projpred site, but this link is currently dead, of course.)
* My questions from #208 have not been answered yet, so for section "Troubleshooting", I was assuming that [this comment](https://github.com/stan-dev/projpred/issues/208#issuecomment-932486314) is correct, leading to the statement
    > For non-Gaussian models, the discrepancy may be due to the fact that the penalized iteratively reweighted least squares (PIRLS) algorithm might have convergence issues (Catalina et al., 2021). In this case, the latent-space approach by Catalina et al. (2021) might help.
    
    in the new vignette.


fweber144:
This fixes #137. Another motivation for this was to speed up the building of the vignette (so it can be checked more often) and to ensure that the vignette on CRAN contains the R output.

fweber144:
Based on personal communication with @avehtari, I added notes that the implementation for additive models is currently only experimental. Corresponding notes at other places of the package (docs, warning in `init_refmodel()`) will be added by another PR.

fweber144:
@jgabry and @paul-buerkner: In the personal communication, @avehtari also asked why I removed the `children/SETTINGS-*` files. He said they were meant to be consistent across Stan-related R packages. I wasn't aware of that and removed them in **projpred** because 2 of the formerly 3 files weren't in use (`SETTINGS-gg.txt` and `SETTINGS-rstan.txt`), so I thought they were remnants of older vignettes. And I included the content from the remaining one (`SETTINGS-knitr.txt`) in the .Rmd file of the vignette itself because there was only a single vignette left. Later, I also slightly modified that setup chunk which used to be in `SETTINGS-knitr.txt` (for example, I altered the figure settings to be more appropriate for this vignette), so now it wouldn't be consistent with other Stan-related packages anymore, even if it was moved back to its own file.

After checking other Stan-related packages, I saw that the `children/SETTINGS-*` files aren't consistent anyway. For example, the **posterior** package doesn't have these files (see [here](https://github.com/stan-dev/posterior/tree/master/vignettes)), **cmdstanr** only has a single file (see [here](https://github.com/stan-dev/cmdstanr/tree/master/vignettes/children)), and bayesplot, too (see [here](https://github.com/stan-dev/bayesplot/tree/master/vignettes/children)). **rstanarm** has 2 files (see [here](https://github.com/stan-dev/rstanarm/tree/master/vignettes/children)). So Aki said we should ask you about these files.

jgabry:
I think `SETTINGS-knitr` is probably the only one to consider using in projpred (it helps certain things in the vignettes look consistent across packages).The others are more specific to individual packages I think.

fweber144:
Ok, so @avehtari do you want me to move the setup chunk back to `SETTINGS-knitr`? And use the figure settings from most other Stan-related R packages? I would prefer not to use `comment = NA` because we have a lot of output in the projpred vignette and with the default `comment` symbols (see [here](https://yihui.org/knitr/options/#code-decoration)), it is easier to distinguish from the code (which is more important for the users).

vignettes/projpred.Rmd
```diff
@@ -0,0 +1,278 @@
...
+```{r, cache = TRUE, cache.lazy = FALSE}
+# Set this manually if desired:
+ncores <- parallel::detectCores(logical = FALSE)
+### Only for technical reasons in this vignette (you can omit this when running
+### the code yourself):
+ncores <- min(ncores, 2L)
+###
+refm_fit <- stan_glm(
+  y ~ X1 + X2 + X3 + X4 + X5 + X6 + X7 + X8 + X9 + X10 + X11 + X12 + X13 + X14 +
+    X15 + X16 + X17 + X18 + X19 + X20,
```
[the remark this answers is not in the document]
fweber144: > `y ~ .` did use to work, but now I get an error for that

Hm, for me it works on branch `develop`. Could you give an example which fails?

avehtari:
I think it would be good to have the settings which will be shared with different projpred vignettes in one place, and checking that the figure settings are similar with other packages is also good. I would use `comment = NA` if it's used by other packages (that's what I also use for all my case studies).

fweber144:
Alright, I just pushed commit 2c87ba4 which does this. Note that the figures are smaller now (this was the reason why I had optimized the figure settings in my custom settings before).

I also pushed commit f9ac051 which introduces a TOC (analogously to **cmdstanr** and **bayesplot**).

avehtari:
I tried to run the vignette, but I get the following error with develop branch
```
Quitting from lines 227-229 (projpred.Rmd) 
Error in predict.subfit(fit, newdata = newdata, weights = weights) : 
  unused argument (weights = weights)
Calls: <Anonymous> ... proj_linpred -> proj_helper -> lapply -> FUN -> <Anonymous>
Execution halted
```
and following error with master branch
```
Quitting from lines 227-229 (projpred.Rmd) 
Error in eval(formula[[2]], data, environment(formula)) : 
  object 'y' not found
Calls: <Anonymous> ... eval2 -> eval -> eval -> .fun -> eval_rhs -> eval -> eval
Execution halted
```
as this PR is only about the vignettes, I assume they should work with develop branch which is the target of the PR.
Am I missing something?

fweber144:
That's strange. You're absolutely right with

> as this PR is only about the vignettes, I assume they should work with develop branch which is the target of the PR.

For me, it works with both, the package at the state of `develop` and the package at the state of this PR. Are you sure your `develop` branch is up-to-date? The latest commit on `develop` should be ef3cfc2 (the merge of PR #235). And then, did you make sure to install projpred from `develop`? The vignette needs the installed package.

avehtari:
> Are you sure your develop branch is up-to-date? The latest commit on develop should be ef3cfc2 (the merge of PR #235). And then, did you make sure to install projpred from develop? The vignette needs the installed package.

I installed with
```
remotes::install_github("stan-dev/projpred", ref = "develop")
```

fweber144:
Ok, that should be fine. But even then, with installation via `remotes::install_github()`, it works for me. Could you maybe post your session info?

fweber144:
For the session info, running
```r
library(rstan)
library(rstanarm)
library(projpred)
library(posterior)
library(bayesplot)
sessionInfo()
```
should be enough.

avehtari:
It works now after I updated all the packages, and not just the packages that are listed as projpred dependencies (ie those which R suggests updating when running `remotes::install_github("stan-dev/projpred", ref = "develop")`)

fweber144:
Ok, thanks for reporting.

fweber144:
Perhaps that was also the reason for the `y ~ .` thing [here](https://github.com/stan-dev/projpred/pull/237#discussion_r743533938)?

vignettes/projpred.Rmd
```diff
@@ -0,0 +1,320 @@
...
+```{r}
+if (!identical(.Platform$OS.type, "windows")) {
+  trigger_default <- options(projpred.prll_prj_trigger = 200)
+  library(doParallel)
+  registerDoParallel(ncores)
+}
+```
+
+It is worth taking a look at the documentation of `cv_varsel()` because we cannot explain all arguments here. By modifying argument `cv_method`, we use a K-fold CV instead of a leave-one-out (LOO) CV. We do this here only to make this vignette build faster. In general, this is not recommended since LOO CV is more accurate. By modifying argument `nterms_max`, we impose a limit on the model size until which the search is continued. Typically, one has to run the variable selection with a large `nterms_max` first (the default value may not even be large enough) and only after inspecting the results from this first run, one is able to set a reasonable `nterms_max` in subsequent runs. The value we are using here (`12`) is based on such a first run (which is not shown here, though).
```
avehtari, line 106: "It is worth taking a look at the documentation of `cv_varsel()` because we cannot explain all arguments here."
->
Here, we use only some of the available arguments; see the documentation of `cv_varsel()` for the full list of arguments.

avehtari: "In general, this is not recommended since LOO CV is more accurate."
->
When the computation time is not an issue, we recommend using LOO-CV as it is likely to be more accurate.

vignettes/projpred.Rmd
```diff
@@ -0,0 +1,320 @@
...
+<!-- Alternative, as soon as GitHub issue #168 is resolved: -->
+<!-- ```{r} -->
+<!-- prj <- project( -->
+<!--   cvvs, -->
+<!--   nterms = modsize_decided, -->
+<!--   cv_search = FALSE, -->
+<!--   seed = 15705533 -->
+<!-- ) -->
+<!-- ``` -->
+For more accurate results, we could have increased argument `ndraws` of `project()` (up to the number of posterior draws in the reference model). This increases the runtime, though, which we don't want in this vignette.
```
avehtari, line 170: "This increases the runtime, though, which we don't want in this vignette."
->
This increases the runtime, which we don't want in this vignette.

vignettes/projpred.Rmd
```diff
@@ -0,0 +1,320 @@
...
+
+A visualization of the projected posterior can be achieved with the **bayesplot** package, for example using its `mcmc_intervals()` function.
+```{r}
+library(bayesplot)
+bayesplot_theme_set(ggplot2::theme_bw())
+mcmc_intervals(prj_mat) +
+  ggplot2::coord_cartesian(xlim = c(-1.5, 1.6))
+```
+
+Note that we only visualize the *1-dimensional* marginals of the projected posterior here. To gain a more complete picture, we would have to visualize at least some *2-dimensional* marginals of the projected posterior (i.e., marginals for pairs of parameters), giving 3-dimensional plots.
```
avehtari, line 198: ", giving 3-dimensional plots."
Drop this (I don't think it makes sense)

vignettes/projpred.Rmd
```diff
@@ -0,0 +1,320 @@
...
+pars_nms_refm <- c(
+  "(Intercept)",
+  sub("^b_", "", grep("^b_X", colnames(prj_mat), value = TRUE)),
+  "sigma"
+)
+mcmc_intervals(refm_mat, pars = pars_nms_refm) +
+  ggplot2::coord_cartesian(xlim = c(-1.5, 1.6))
+```
+
+Here, the reference model's marginal posteriors differ only slightly from the marginals of the projected posterior. This does not necessarily have to be the case, though.
```
avehtari, line 212: ", though"
drop this

avehtari (changes requested):
I added some comments to improve the text. After those changes, this is ready to be merged.

fweber144:
Thanks, done (see commit 56e05e4).

avehtari:
Thanks for the patience!

fweber144:
No problem. Thanks for the review and the merge!

PR #254, opened 2021-12-10
Docs, vignette, `dis`, `predvar`, `kl`

fweber144:
This is a collection of several, rather minor (no-op) changes, in particular concerning:
* the documentation
* the vignette
* the dispersion parameter (`dis`)
* the predictive variances (`predvar`)
* the KL divergence (`kl`)

See the commit messages for details.


fweber144:
Sorry, I made a mistake in the comment of commit 47138e7: There, I said

> For `datafit`s (but only for those where a dispersion parameter exists), keep
using 0 instead of NA in case of an unspecified `dis`. The reason is that when
using NAs, the NAs propagate down to the results (e.g., `summary()`'s output) and
furthermore, 0 indeed makes sense for `datafit`s because there, we have
`identical(as.vector(<datafit_object>$mu), <datafit_object>$y)` giving `TRUE`.

However, `identical(as.vector(<datafit_object>$mu), <datafit_object>$y)` giving `TRUE` does *not* mean that there's no uncertainty due to the observational model (the Gaussian distribution of the response values, for example). So I will create a PR where `NA` is also used as the "defunct" value for `datafit`s.

PR #260, opened 2021-12-12
Refactors due to augmented-data projection

fweber144:
While working on the yet-to-come PR for the augmented-data projection, I performed some refactors also in the not-yet extended code. These are collected here. See the commit messages for details.


fweber144:
Sorry, had to force-push because the most recent commit e9da36236e45a01a8ab6a7b490e60f94fb1d851c will be part of another PR.

Issue #154, opened 2021-06-20
Name and documentation of `cv_search`


fweber144:
I wonder whether the name and the documentation of `project()`'s, `varsel()`'s, and `cv_varsel()`'s argument `cv_search` makes sense.

The documentation from `project()` says about `cv_search`:

> If TRUE, then the projected coefficients after L1-selection are computed without any penalization (or using only the regularization determined by regul). If FALSE, then the coefficients are the solution from the L1-penalized projection. This option is relevant only if L1-search was used. Default is TRUE for genuine reference models and FALSE if object is datafit (see init_refmodel).

However, when looking at the source code of `.get_submodels()` (to which `project()`, `varsel()`, and `cv_varsel()` pass their argument `cv_search`), `cv_search` seems to be relevant for non-L1 search, too. I guess the comment from [these lines](https://github.com/stan-dev/projpred/blob/dc1c3e473c8df2298e956c406499ddac941e5c5f/R/projfun.R#L50-L54) might be more up-to-date than the documentation. If that comment is correct, I think `cv_search` should be renamed to something like `reproject` (or `refit_proj` which occurs on branch `workflow` and seems to be related to `cv_search`?).

jpiironen:
This documentation item you copy-pasted above is written by me a long time ago, but it is actually for an input argument that at that time was called `relax`. I don't know what has happened, apparently the `relax` argument has been removed at some point (@AlejandroCatalina ?) and the documentation has gotten corrupted so that the documentation for `cv_search` is completely wrong. 

I checked out some very old commit, and it seems at that time the `cv_search` was called `validate_search` and the corresponding documentation item was: 
```
Whether to cross-validate also the selection process, that is, whether to perform
selection separately for each fold. Default is TRUE and we strongly recommend not setting this
to FALSE, because this is known to bias the accuracy estimates for the selected submodels.
However, setting this to FALSE can sometimes be useful because comparing the results to the case
where this parameter is TRUE gives idea how strongly the feature selection is (over)fitted to the
data (the difference corresponds to the search degrees of freedom or the effective number
of parameters introduced by the selection process).
```

By the way, I don't quite get why `cv_search` is also an argument to `varsel`, because `varsel` does not perform cross-validation. It should only be an argument to `cv_varsel`.

AlejandroCatalina:
[...]

Issue #211, opened 2021-09-24
Bug in `project()$p_type`


fweber144:
In case of `cv_search = FALSE`, line https://github.com/stan-dev/projpred/blob/9c7fd061f078e8d4f410f86e377362d43908eac0/R/project.R#L230 can give an incorrect result:
```r
options(mc.cores = parallel::detectCores(logical = FALSE))
data("df_gaussian", package = "projpred")
mydat <- cbind("y" = df_gaussian$y, as.data.frame(df_gaussian$x))
library(rstanarm)
myfit <- stan_glm(y ~ V1 + V2 + V3 + V4 + V5,
                  data = mydat,
                  seed = 1140350788)

library(projpred)
my_vs <- varsel(myfit, ndraws = 21, ndraws_pred = 25, nterms_max = 2)
myprj <- project(my_vs, nterms = 1, cv_search = FALSE)
myprj$p_type
## --> FALSE (correct)
myprj_wrong <- project(my_vs, nterms = 1, cv_search = FALSE, nclusters = 2)
myprj_wrong$p_type
## --> TRUE (incorrect)
```
This bug is important for the warning thrown by `as.matrix.projection()`, for example.

I think the best solution would be to use
```r
p_used <- if (cv_search) p_ref else object$search_path$p_sel
```
before creating output object `proj` and then something like
```r
model$p_type <- length(unique(p_used$weights)) != 1
```
when creating output object `proj`. Then however, the docs for the return value of `project()` (more precisely, only element `p_type`) would have to be adapted (because then, `p_type` does not always indicate whether clustering was used; only if the (clustered) projected draws have different weights). The warning https://github.com/stan-dev/projpred/blob/9c7fd061f078e8d4f410f86e377362d43908eac0/R/methods.R#L1025-L1028 would have to be changed to something like
```r
warning("The reference model's posterior draws have been clustered for ",
        "the projection and the clusters have different weights. ",
        "Therefore, you should not use the results from this as.matrix() ",
        "method without taking the weights into account.")
```
The documentation for argument `nresample_clusters` would also have to be adapted (because then, `nresample_clusters` is not always used in case of clustering; only if the clusters have different weights). However, since I would suggest to remove the resampling of clusters anyway and let the `posterior` package do that (see issue #206), I think it's better to solve issue #206 first.

In any case, at that occasion, I think we should rename `p_type` to something more meaningful, like `has_uneq_wdraws`.

fweber144:
[...]

Issue #168, opened 2021-07-08
Re-use of fitted submodels


fweber144:
Argument `cv_search` currently seems to be be responsible for fetching submodels already fitted during a search (see issue #154 and function `.get_submodels()`). The relevant code seems to be line https://github.com/stan-dev/projpred/blob/16770099ae45c24a3e9b1798ec46c36218314a68/R/projfun.R#L66 There however, the fitted submodels from `search_path` are fetched which usually have a coarser reference distribution (`p_sel`) than used for prediction (`p_pred`), see lines https://github.com/stan-dev/projpred/blob/16770099ae45c24a3e9b1798ec46c36218314a68/R/varsel.R#L96-L98 and https://github.com/stan-dev/projpred/blob/16770099ae45c24a3e9b1798ec46c36218314a68/R/cv_varsel.R#L292-L304 and https://github.com/stan-dev/projpred/blob/16770099ae45c24a3e9b1798ec46c36218314a68/R/cv_varsel.R#L558-L559 where `ndraws` and `nclusters` are usually smaller than their `_pred` variants. So my question is: Wouldn't it make more sense to let `cv_search` fetch the fitted submodels which are based on `p_pred`, i.e., those which were used for prediction? Of course, that would only be possible when using `cv_search` in the context of `project()` (not within `varsel()` or `cv_varsel()`). In order to be able to do this, I guess `.get_submodels()` would need a new (optional) argument accepting the fitted submodels based on `p_pred`. Furthermore, `p_sub` from https://github.com/stan-dev/projpred/blob/16770099ae45c24a3e9b1798ec46c36218314a68/R/varsel.R#L110-L113 would need to be added to the respective `"vsel"` output object (which is probably helpful anyway) of `varsel()` and `cv_varsel()`. Then `project()` could pass that `p_sub` component of the `"vsel"` object to `.get_submodels()`.

AlejandroCatalina:
So the idea of this argument is to avoid computational cost by not refitting the projections with `ndraws_pred` precisely. We can add your suggestion for `project` only. By the way, this argument will go back to being named `relax` probably, as it matches its purpose properly.

fweber144:
> So the idea of this argument is to avoid computational cost by not refitting the projections with ndraws_pred precisely.

Yes, my proposition would also avoid refitting with `ndraws_pred` as it would re-use the corresponding fits from `varsel()` or `cv_varsel()` in `project()`.

> We can add your suggestion for project only.

[...]

PR #265, opened 2022-01-12
Fix issue #154

fweber144:
This fixes issue #154.


fweber144:
With this renaming of argument `cv_search`, we could also test this argument more thoroughly than we are doing now. I haven't done this yet because I think issues #211 and #168 should be solved first.

Issue #156, opened 2021-06-21
Documentation of `formula` for additive models


fweber144:
In the documentation for argument `formula` of `get_refmodel.default()` and `init_refmodel()`, I think some additional information about additive models would be helpful (or at least where to find it). I'm not too familiar with the projpred implementation for additive models, so I guess you (the projpred authors) can add that information more easily.

fweber144:
Might be fixed in `develop` now (I'm not sure if my added documentation is sufficient).

fweber144:
Just stumbled across your comment

> This happens because we have not implemented parsing support for by syntax in smooth terms. It is desirable to have but we haven’t had the time yet.

from #145. Are you sure the group-level independent smoothers (`s(x, by = g)`) are supported?

AlejandroCatalina:
No, that’s what I said, it is not implemented, or at least not properly .

Best,
Alex
________________________________

fweber144:
Ok, then I probably got you wrong in this thread here.

fweber144:
There are several things I'd like to point out here:

First of all, to see why the `split_formula()` thing is problematic, use `varsel()` instead of `project()`:
```r
data("lasrosas.corn", package = "agridat")
lasrosas.corn$year <- as.factor(lasrosas.corn$year)
library(rstanarm)
options(mc.cores = parallel::detectCores(logical = FALSE))
refm_fit <- stan_gamm4(
  yield ~ year + topo + s(bv, by = topo),
  family = gaussian(),
  data = lasrosas.corn,
  seed = 4919670
)
library(projpred)
vs <- varsel(refm_fit, ndraws_pred = 20, seed = 15705533)
```
which gives
```
Error in class(ff) <- "formula" : cannot set attribute on a symbol
In addition: Warning messages:
1: In init_refmodel(object = object, data = data, formula = formula,  :
  Support for additive models is still experimental.
```

Secondly, the same thing happens with `t2()` instead of `s()` (here demonstrated with two predictors in the `t2()` call):
```r
refm_fit <- stan_gamm4(
  yield ~ year + topo + t2(bv, nitro, by = topo),
  family = gaussian(),
  data = lasrosas.corn,
  seed = 4919670
)
prj <- project(
  refm_fit,
  solution_terms = c("topo + t2(bv, nitro, by = topo)", "year"),
  ndraws = 25,
  seed = 15705533
)
vs <- varsel(refm_fit, ndraws_pred = 20, seed = 15705533)
```

[...]

PR #269, opened 2022-01-24
Fix issue #156

fweber144:
This fixes issue #156.

### DIFF
--- a/R/formula.R
+++ b/R/formula.R
@@ -84,14 +84,8 @@ extract_response <- function(response) {
   return(response_name_ch)
 }
 
-## Parse additive terms from a list of individual terms. These may include
-## individual smoothers s(x), interaction smoothers s(x, z), smooth intercepts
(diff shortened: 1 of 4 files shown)


fweber144:
Fixed in `develop`, so a corresponding label may be added here.

fweber144:
Thanks for merging this. Just for the sake of completeness: I forgot to mention the following remarks which might become relevant when support for additive models is extended:

* I merged the "population-level interaction smoothers" into the "interaction smoothers" because to me, they appeared to be the same.
* In the `t2()` (formerly `te()`) term, I omitted argument `bs` because of what is described in issue #156. Btw, I think this argument `bs` can be omitted in the demonstration anyhow because its default is `"cr"` and that should be the typical use case. (Compared to cubic regression splines, thin plate regression splines don't have an advantage as marginal bases for tensor product smooths (see `?mgcv::smooth.terms`) and they are computationally more expensive.)

PR #270, opened 2022-01-25
Collection of minor improvements

fweber144:
These are multiple minor improvements. Sorry that I had to collect them into one PR again, but I couldn't find a sensible other way for submitting these changes (one PR per minor change would certainly have been overkill). I guess you need to check the commits here one-by-one to see if you want to accept them. Details may be found in the commit messages.


fweber144:
I just pushed some more commits (b07bf3e to 01682fb). I hope this is ok.

PR #271, opened 2022-02-09
Add `cvrefbuilder`

fweber144:
This PR is necessary in conjunction with PRs paul-buerkner/brms#1287 and paul-buerkner/brms#1290 to solve paul-buerkner/brms#1286. But as mentioned in the message of commit 09c7286c0fb261a13bf0e29ff0e30b745a14a98c, this is also necessary for some custom reference models.


fweber144:
Sorry, I noticed an incorrect cross-link in the message of former commit 8148008c (now b76ceff) and therefore had to force-push to achieve the reword.

Issue #152, opened 2021-06-11
`as.matrix.projection()` for GAMs: Naming scheme of parameters


fweber144:
For GAMs, `as.matrix.projection()` does not follow brms's naming scheme (discussed in #75):

On branch `develop` (commit 96d8e11d6f0ad0a1f7836c51700cb676b364bee5), this reprex:
```r
options(mc.cores = parallel::detectCores(logical = FALSE))
library(brms)
data(kidiq, package = "rstanarm")
fit_gauss <- brm(formula = kid_score ~ s(mom_iq),
                 data = kidiq,
                 seed = 469230)

library(projpred)
prj <- project(fit_gauss, solution_terms = c("s(mom_iq)"),
               ndraws = 25)
colnames(as.matrix(prj))
```
gives
```
 [1] "b_Intercept"   "b_s(mom_iq).1" "b_s(mom_iq).2" "b_s(mom_iq).3"
 [5] "b_s(mom_iq).4" "b_s(mom_iq).5" "b_s(mom_iq).6" "b_s(mom_iq).7"
 [9] "b_s(mom_iq).8" "b_s(mom_iq).9" "sigma"
```
whereas
```r
colnames(as.matrix(fit_gauss))
```
gives
```
 [1] "b_Intercept"    "bs_smom_iq_1"   "sds_smom_iq_1"  "sigma"         
 [5] "s_smom_iq_1[1]" "s_smom_iq_1[2]" "s_smom_iq_1[3]" "s_smom_iq_1[4]"
 [9] "s_smom_iq_1[5]" "s_smom_iq_1[6]" "s_smom_iq_1[7]" "s_smom_iq_1[8]"
[13] "lp__"
```

Just a side note: As may be seen in issues #150 and #151, rstanarm uses a different naming scheme. It might make sense to have separate naming schemes depending on whether the reference model inherits from a `"stanreg"`, a `"brmsfit"`, or some different object. (In the latter case, one could use either rstanarm's or brms's naming scheme.) But that's probably a larger change and may be deferred to the future. I think for now, using brms naming scheme consistently should be enough.

PR #279, opened 2022-02-16
Naming scheme of `as.matrix.projection()`'s output columns

fweber144:
This makes the scheme for naming the columns (i.e., the parameters) of `as.matrix.projection()`'s output matrix dependent on the package used for fitting the reference model: For brms reference models, brms's naming scheme is used; for rstanarm reference models, rstanarm's naming scheme is used. If the reference model fit is of unknown class, rstanarm's naming scheme is used. The user can also set the naming scheme manually, using `as.matrix.projection()`'s new argument `nm_scheme`.

This was already mentioned in [this comment in issue #152](https://github.com/stan-dev/projpred/issues/152#issue-918920675) and was now requested by @avehtari.

### DIFF
--- a/R/methods.R
+++ b/R/methods.R
@@ -803,17 +803,21 @@ suggest_size.vsel <- function(
   return(suggested_size - 1) ## substract the intercept
 }
 
-replace_intercept_name <- function(nms, nm_scheme) {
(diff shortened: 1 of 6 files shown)


avehtari:
Thanks!

Issue #268, opened 2022-01-23
Prediction for new group levels (rstanarm vs. lme4)


fweber144:
For multilevel rstanarm reference models, there is an inconsistency between the reference model and the submodels with respect to the calculation of the linear predictor for new group levels in a K-fold CV: rstanarm reference models call `rstanarm::posterior_linpred.stanreg()` with no other arguments apart from `object` and `newdata`: https://github.com/stan-dev/projpred/blob/d7c64e469ff986a26dc872e8a12f6119bdba97dd/R/refmodel.R#L498 Thus, according to rstanarm's docs (see also the example below), this call marginalizes over the group-level effects (only for the new levels). In contrast, for the submodels, `lme4::predict.merMod()` is called with no other arguments apart from `object`, `newdata`, and `allow.new.levels = TRUE`: https://github.com/stan-dev/projpred/blob/d7c64e469ff986a26dc872e8a12f6119bdba97dd/R/divergence_minimizers.R#L427 Thus, according to lme4's docs (see also the example below), this call sets group-level effects for new levels to zero.

My question is now: Do you think this needs to be changed and if yes, how? The problem is that rstanarm's options for dealing with new levels are not quite the same as lme4's options. The only common possibility (as far as I understand these two packages) is to set group-level effects for *all* levels (i.e., existing and new ones) to zero. This, however, has important implications for the evaluation of the models' predictive performance: Adding a group-level effect to a submodel which previously only included population-level effects would then make no (or in case there is a dispersion parameter, only little) difference with respect to the predictive performance measures like ELPD, RMSE, etc. In other words, due to the existing levels, the value of the performance measures might look worse than predictive performance actually is. Apart from that, I'm not sure if, for families with a dispersion parameter, the original dispersion parameter estimates would then still be appropriate (similarly to [this](https://discourse.mc-stan.org/t/bayes-r2-and-conditioning-on-random-effects/20461/20?u=fweber144)).

Needed for the following examples:
```r
# Data --------------------------------------------------------------------

set.seed(457211)
dat <- matrix(rnorm(41 * 2, sd = 0.4),
              ncol = 2,
              dimnames = list(NULL, paste0("X", 1:2)))
dat <- data.frame(dat)
dat$group <- gl(n = 8, k = floor(nrow(dat) / 8), length = nrow(dat),
                labels = paste0("gr", seq_len(8)))
group_icpts_truth <- rnorm(nlevels(dat$group), sd = 6)
icpt <- -4.2
[...]

PR #287, opened 2022-03-11
Fix issue #268

fweber144:
This fixes issue #268 for GLMMs by drawing the random effects for new group levels from a (multivariate) Gaussian distribution.

Note that by this PR, issue #268 is not fixed for GAMMs yet.

**Copyright/license:**

The code for the new `repair_re.merMod()` is inspired by (i.e., it uses code snippets from) `lme4:::predict.merMod()`. I included a corresponding statement right before `repair_re.merMod()` ([these lines](https://github.com/fweber144/projpred/blob/5a32fc6b17dd67aa6dfe712d4a655dbfb74b92bf/R/divergence_minimizers.R#L501-L514)). Since **lme4** is licensed under GPL (>=2) and **projpred** is licensed under GPL-3, the inclusion of code snippets from **lme4** should be ok. (For a proof that GPL (>=2) and GPL-3 are compatible, see [here](https://www.gnu.org/licenses/gpl-faq.html#v2v3Compatibility) and the table [here](https://www.gnu.org/licenses/gpl-faq.html#AllCompatibility).) However, some remaining questions I have are:

1. Do you want me to add a separate `LICENSE` file where we explain which code snippets from **lme4** are used (and change the "License" field in `DESCRIPTION` correspondingly to `GPL-3 | file LICENSE`)? This is what is done in package [**shiny**](https://CRAN.R-project.org/package=shiny), for example. (But I guess the external components in **shiny** are larger than the code snippets used here.)
2. At that occasion (of enumeration point 1), do you also want me to add a `LICENSE.md` file containing the full GPL-3 license text (see [here](https://r-pkgs.org/license.html#key-files))?
3. @bbolker (and all other **lme4** authors): Are [these lines](https://github.com/fweber144/projpred/blob/5a32fc6b17dd67aa6dfe712d4a655dbfb74b92bf/R/divergence_minimizers.R#L501-L514) a sufficient copyright/licensing approach for you and the other **lme4** authors? Or do you prefer a copyright-holder entry in the "Authors@R" field of the `DESCRIPTION` file? If you would like such a copyright-holder entry in the "Authors@R" field, which persons should it include? Or simply "The LME4 Authors"?

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -2,6 +2,10 @@
 ^.*\.Rproj$
 ^\.Rproj\.user$
 ^\.github$
+^LICENSE\.md$
+^README\.Rmd$
+^cran-comments\.md$
+^CRAN-RELEASE$
 ^man-roxygen$
 ^tests/testthat/bfits$
 ^tests/testthat/_snaps$
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -24,7 +24,7 @@ Description:
     'rstanarm' and 'brms' packages, but other reference models can also be used.
     See the documentation as well as the package vignette for more information
     and examples.
-License: GPL-3
+License: GPL-3 | file LICENSE
 URL: https://mc-stan.org/projpred/, https://discourse.mc-stan.org
 BugReports: https://github.com/stan-dev/projpred/issues/
 Depends:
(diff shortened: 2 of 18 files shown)


avehtari:
I think your suggestions make sense

bbolker:
I think what you've done here is fine, thanks for asking (I'll forward this to other lme4-authors in case they have strong opinions, but I'm pretty sure they'll be fine with this too).

fweber144:
> I think what you've done here is fine, thanks for asking (I'll forward this to other lme4-authors in case they have strong opinions, but I'm pretty sure they'll be fine with this too).

Great, thanks.

PR #305, opened 2022-04-23
Vignette: Use LOO CV with `validate_search = FALSE` instead of K-fold CV

fweber144:
This makes the vignette use `cv_varsel()` with LOO CV and `validate_search = FALSE` instead of K-fold CV. This provides a speed improvement when building the vignette and should reflect the typical projpred workflow better (running `cv_varsel()` with `validate_search = FALSE` first for rough results and only later running `cv_varsel()` with `validate_search = TRUE`). Of course, this causes the vignette to show an unvalidated search, but this is emphasized in the text, so I think that should be ok.

@avehtari: Do you want to take a look at this? You don't need to, but you contributed some helpful comments on the last vignette PR (#237). I added you as reviewer, but could remove you again.


avehtari:
After running `cv_varsel` with `validate_search = FALSE` I would add  something like

> we directly plot with `deltas = TRUE` here (note that as the `validate_search = FALSE` this result is slightly optimistic, and the plot looks different when `validate_search = TRUE` is used):

> close enough to the reference model's performance (this is also affected by `validate_search = FALSE`)

> close enough to the reference model's performance (this is also affected by `validate_search = FALSE`)

> same final submodel size (`6`) as by our manual decision (`suggest_size()` is also affected by `validate_search = FALSE`)

PR #322, opened 2022-05-14
Augmented-data projection

fweber144:
This PR adds support for the augmented-data projection to solve #70. For now, I consider it as experimental because

1. I still need to add more thorough unit tests for the newly supported models (families `brms::cumulative()` and `brms::categorical()`, both with and without multilevel terms, as well as `rstanarm::stan_polr()` fits which cannot feature multilevel terms);
2. the augmented-data projection needs to be investigated more thoroughly from a theoretical point of view. 

Because of these unfinished aspects, I put "core part" in the PR title. Despite this experimental state, I thought it would not harm to create this PR now so it can already be reviewed and so users can test the augmented-data projection before it is released on CRAN.

**Copyright/license:**

Similarly to PR #287, the code for the new `repair_re()` methods introduced by this PR is inspired by (i.e., it uses code snippets from) `lme4:::predict.merMod()`. I included corresponding comments in the code right before the definitions of these methods and updated the `LICENSE` file correspondingly. However, this PR also introduces the internal `mkNewReTrms_man()` function which is basically a copy of `lme4::mkNewReTrms()`. I have added a corresponding entry in the `LICENSE` file and (since we now basically have a copy of **lme4** code in **projpred**) also added the **lme4** authors as copyright holders to the `Authors@R` field in the `DESCRIPTION` file. So the question to @bbolker (and all other **lme4** authors) is if you are ok with being listed as copyright holders in the `Authors@R` field and if that entry is ok this way (see [these lines](https://github.com/fweber144/projpred/blob/2c140d3cafe51a4936f8a77ad11ff42324b8dfa3/DESCRIPTION#L18-L19)).

**EDIT 1:** Point 1 above (thorough unit tests) has been accomplished now (current status: commit 5fe5839).

**EDIT 2:** Point 2 above (theoretical investigations) has been accomplished now (https://doi.org/10.48550/arXiv.2301.01660). Thus, with commit 89f0baa, the augmented-data projection is not considered as an experimental feature anymore.


bbolker:
Sorry, this slipped through the cracks. I am fine with this, and I'm 99% sure that Martin Maechler and Doug Bates are all as well (they're all staunch free-software people, and this is all GPL-3, so really you can use this any way you want ...) I will pass this along to them, but I think you're fine.

avehtari (approved):
Read through all docs and did not find any errors or strane things

fweber144:
> Read through all docs and did not find any errors or strane things

Thanks

Issue #318, opened 2022-05-10
AUC upper/lower quantiles a bit too approximate


awd97:
We prefer to show both elpd and auc results for our logistic regressions, to placate the non-statistician readership of our work (!)

When outputting auc lower and upper quantiles using summary.vsel, it sometimes gives upper quantiles above 1. From the code it looks like the se is computed as the sd of the bootstrap samples, and then this se is used to calculate the lower and upper quantiles assuming normal distribution around the mean, which it is not. Could you calculate the quantiles directly from the bootstrap samples instead?

I have forked projpred and could issue a pull request.

awd97:
no idea how this came to be a duplicate of #317...

fweber144:
Yes, you are right that projpred uses a normal-approximation interval for the performance statistics, even in case of bootstrapping. If I understand you correctly, you propose to use the interval obtained (for example, could also be calculated manually) by `type = "perc"` in `boot::boot.ci()`. That sounds reasonable and I would also favor this compared to the normal-approximation interval. I haven't had the time yet to continue work on the bootstrap (issue #216 is still open). If you could submit a PR, that would be great!

fweber144:
Resolved by #347.

Issue #94, opened 2021-03-19
LOO subsampling weights in cv_varsel.refmodel()


fweber144:
In `cv_varsel.refmodel()`, the observation-specific weights are hard-coded to be all equal (line https://github.com/stan-dev/projpred/blob/0cabd73548eec32e9b675f15cd39d377a40f180a/R/cv_varsel.R#L221 ), but in case of subsampling LOO-CV, they might be different (as the comment says there, too). Thus, shouldn't that line be modified?

AlejandroCatalina:
Yes. This is a bit less straightforward to fix as we can compute both approximate LOO or full LOO, so I think the right weights should be returned by `loo_varsel` and `kfold_varsel` themselves. I'll work on this after resolving the easier issues.

fweber144:
Related comments/discussions may be found in issue #173 (enumeration point 4) and PR #188. Now that I understand the purpose of `w` better, I think constant weights for LOO-CV are ok (no matter whether subsampled or not) but they might be inappropriate for K-fold CV with folds of very different sizes.

fweber144:
I have now read Magnusson et al. (2019) and Magnusson et al. (2020), but I'm still confused how **projpred** wants to perform the subsampled LOO CV. More importantly, I have the strong feeling that the current implementation is not correct. What I learned from Magnusson et al. (2019) and Magnusson et al. (2020) makes me think the problems with the implementation in **projpred** get deeper than I initially thought.

What is important for **projpred** is that:

1. Magnusson et al. (2019) propose to perform probability-proportional-to-size sampling (PPS) combined with the Hansen-Hurwitz (HH) estimator.
1. Magnusson et al. (2020) propose to perform simple random sampling (SRS) combined with the difference estimator. The reason for this new proposal is that the approach by Magnusson et al. (2019) is not optimal for model comparisons (as explained by Magnusson et al., 2020).

[...]

PR #347, opened 2022-09-07
Update summary_funs.R

VisionResearchBlog:
This follows up from @awd97 [thread here](https://github.com/stan-dev/projpred/issues/318) where current AUC calculation can be improved by using the lower/upper quantile computation from bootstrap. The original correction was for 2.1.0 code base, this updates to 2.2.0

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -14,7 +14,8 @@ Authors@R: c(person("Juho", "Piironen", role = c("aut"),
              person("Jonah", "Gabry", role = "ctb"),
              person("Marco", "Colombo", role = "ctb"),
              person("Paul-Christian", "Bürkner", role = "ctb"),
-             person("Hamada S.", "Badr", role = "ctb"))
+             person("Hamada S.", "Badr", role = "ctb"),
+             person("Brian", "Sullivan", role = "ctb"))
 Description:
     Performs projection predictive feature selection for generalized linear and
     additive models as well as for generalized linear and additive multilevel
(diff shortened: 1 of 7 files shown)


fweber144:
Thanks! I'll take a look at this asap.

fweber144:
Most of this looks good, thanks again for your contribution. I have pushed some commits to this PR branch. Things that need to be done or at least clarified before merging are highlighted by `TODO (bootstrap)` in the code. Currently, this is basically one question: Why didn't you use this kind of bootstrap interval also for the RMSE?

Furthermore, I have added a `TODO (subsampling LOO-CV)` comment. I have still not decided whether we should do this before merging as well, but I think with the warning I have added, it should be ok for now (subsampled LOO-CV is currently an experimental feature anyway, see #94). In any case, this needs to be tackled, be it now or in the future.

I have added you (@VisionResearchBlog) to the list of contributors. Please tell me if this is not ok or needs to be adapted.

fweber144:
> Ah, and the tests might need to be updated. I'll check this.

The tests are fine.

VisionResearchBlog:
Hi @fweber144 thanks for pointing out not aligning RMSE with the new calculation - It was an oversight as our project is focused on AUC performance.

fweber144:
No problem, I will then implement this myself and push here.

> Furthermore, I have added a `TODO (subsampling LOO-CV)` comment. I have still not decided whether we should do this before merging as well, but I think with the warning I have added, it should be ok for now (subsampled LOO-CV is currently an experimental feature anyway, see #94). In any case, this needs to be tackled, be it now or in the future.

I discussed this with @avehtari and he was ok to leave this as-is for now. So after implementing the RMSE thing, we can merge.

Issue #186, opened 2021-08-09
Offset issues in rstanarm


fweber144:
There are two rstanarm issues related to offsets (stan-dev/rstanarm#541 and stan-dev/rstanarm#542) which require a workaround in projpred until they are resolved in rstanarm. I can try to create a PR for these workarounds in projpred, but before doing so, I want to make sure I understand the key idea of handling offsets in projpred: Is it correct that `<refmod_object>$ref_predfun()` is supposed to *not* take the offsets into account (or, equivalently, use offsets which are all equal to zero) so that the submodels (which basically fit to `<refmod_object>$mu` which in turn is based on `<refmod_object>$ref_predfun()`) may be fitted without any offsets?

AlejandroCatalina:
Submodels can be fitted with offsets if I recall correctly.

Best,
Alex
________________________________
From: Frank Weber ***@***.***>
Sent: Monday, August 9, 2021 11:12:57 AM
To: stan-dev/projpred ***@***.***>
Cc: Subscribed ***@***.***>
Subject: [stan-dev/projpred] Offset issues in rstanarm (#186)


There are two rstanarm issues related to offsets (stan-dev/rstanarm#541<https://github.com/stan-dev/rstanarm/issues/541> and stan-dev/rstanarm#542<https://github.com/stan-dev/rstanarm/issues/542>) which require a workaround in projpred until they are resolved in rstanarm. I can try to create a PR for these workarounds in projpred, but before doing so, I want to make sure I understand the key idea of handling offsets in projpred: Is it correct that <refmod_object>$ref_predfun() is supposed to not take the offsets into account (or, equivalently, use offsets which are all equal to zero) so that the submodels (which basically fit to <refmod_object>$mu which in turn is based on <refmod_object>$ref_predfun()) may be fitted without any offsets?

fweber144:
Yes, in principle they should be fittable with offsets (for example, `lme4::lmer()` accepts offsets via argument `offset` or via `offset()` terms in the formula). But the question is if projpred *wants* them to be fitted with or without offsets. Continuing the `lme4::lmer()` example: `fit_glmer_callback()` does not make use of `lme4::lmer()`'s argument `offset`. And as far as I can tell, the submodel formulas exclude `offset()` terms. So at least for GLMMs, the submodels seem to be fitted without offsets. And I'm not 100% sure, but the same seems to hold for GLMs, probably also for GAMs and GAMMs.

[...]

PR #351, opened 2022-09-19
Disallow offsets for `datafit`s

fweber144:
Following https://github.com/stan-dev/projpred/issues/186#issuecomment-1245771190, this PR disallows offsets for `datafit`s because the submodel fitting does not take offsets into account (but `<refmodel>$mu` contains the observed response values which inevitably "include" the offsets, i.e., we cannot get (hypothetical) observed response values excluding offsets if offsets are actually present).


fweber144:
Note that strangely, the test `"L1-projection with data reference gives the same results as Lasso from glmnet."` also succeeded before the changes made by commit 422ffad24a395fd1dd58346072c09fa4b44bb11c. This might have been due to very relaxed expectations.

Issue #345, opened 2022-08-08
search_terms argument in cv_varsel not working properly


sor16:
Hi,

I am interested in doing variable selection on just a subset of variables, while including the rest in all submodels. This naturally leads to the use of the search_terms argument within cv_varsel. Assume I have a data set of 5 covariates, a treatment variable t and a response y. I let one of the 5 covariates be a confounder, so I am interested in doing variable selection of the remaining 4 covariates, while including t and the confounder x1 in each submodel. A reproducible example of the data generating process, reference model construction and cv_varsel call are below:
 ```
library(brms)
library(projpred)
#function to create the space of submodels to consider
get_search_terms <- function(idx_select, idx_fixed = NULL){
  search_terms <- c()
  n_select <- length(idx_select)
  for(i in 1:n_select){
    search_terms <- c(search_terms, apply(combn(idx_select[1:n_select], i), 2, function(x) paste(paste0("x", x),collapse = "+")))
  }
  fixed_terms <- ''
  if(length(idx_fixed)!=0){
    fixed_terms <- paste(paste0('x', idx_fixed), collapse = '+')
    search_terms <- paste(fixed_terms, search_terms,sep = '+')
    base_model <- paste('t',fixed_terms,sep='+')
  }else{
    base_model <- 't'
  }

  search_terms <- c(base_model,paste0('t+', search_terms))
  return(search_terms)
}

set.seed(2)
N <- 100
p <- 5 #number of additional parameters
p_conf <- 1 #number of confounding variables
p_rel <- p_conf + 2 #number of "relevant" parameters affecting y, excluding t
dat <- as.data.frame(matrix(rnorm(N*p), nrow = N, ncol = p)) #initialize data frame with p covariates
names(dat) <- paste0('x', 1:p)

if(p_conf>0){
  t_betas <- rnorm(p_conf) # effects of confounders on treatment
  dat$t <- rnorm(N, as.matrix(dat[, paste0('x', 1:p_conf)]) %*% t_betas) #treatment is a noisy observation of a linear combination of confounders
}else{
  dat$t <- rnorm(N) #no confounders
}

y_betas <- rnorm(p_rel+1) # effects of treatment and "relevant" parameters on response y
dat$y <- rnorm(N, mean=as.matrix(dat[, c('t', paste0('x', 1:p_rel))]) %*% y_betas) #y is a noisy observation of a linear combination of relevant parameters and treatment
[...]

PR #360, opened 2022-10-18
Fix345 sor16

sor16:
This pull request is meant to fix issue #345. It did so by removing the `candidate_terms` vector, whose matching with the `solution_terms` vector was causing the issue, from the code and using the `solution_terms` instead where needed. Additionally, I fixed what I believe is a bug in `select_possible_terms_size` when full formulas are used in the `search_terms` argument. After the bug fix, all variables already selected are removed in the `full_valid_submodels` vector. The corresponding bug fix required a change in one test which anticipated the behaviour the claimed bug contained.


fweber144:
Thanks a lot, great work! Indeed, I guess `candidate_terms` and the code related to it was a remnant from older projpred versions (< 2.0.0) where the solution terms were (or could be) numeric values.

I have pushed some more commits (mostly cleaning and refactoring). Now I need to check some things and then this should be ready for merging.

R/formula.R
```diff
@@ -647,7 +647,8 @@ select_possible_terms_size <- function(chosen, terms, size) {
   valid_submodels <- unlist(valid_submodels[!is.na(valid_submodels)])
   if (length(chosen) > 0) {
     add_chosen <- paste0(" + ", paste(chosen, collapse = "+"))
-    remove_chosen <- paste0(" - ", paste(chosen, collapse = "-"))
+    remove_chosen <- paste0(" - ",
+                            paste(gsub("\\+", "-", chosen), collapse = "-"))
```
fweber144, line 651: As far as I can tell from playing around with the reprex from #345, this change is not necessary for fixing #345. But I'll keep it nonetheless because it comes with an improvement in terms of efficiency, as can be seen from
```r
allterms <- c("x1 + x2", "x1 + x3", "x1 + x2 + x3")
chosen <- c("x1 + x2")
size_chosen <- count_terms_chosen(chosen)
select_possible_terms_size(chosen, allterms, size = size_chosen + 1)
```
which gave
```
[1] "x3 + x2" "x2 + x3"
```
before this change and gives
```
[1] "x3"
```
after this change. Thus, the new behavior will result in fewer models to be fitted (i.e., to be projected onto) within `search_forward()`.

fweber144: And by the way, the former code just seemed like it was unintended.

fweber144: Just for my future me: Related to this `formula.R` change is the `test_varsel.R` change. The `cv_varsel.R` changes are not related to this `formula.R` change (but are indeed required to fix #345).

fweber144:
All tests passed, so I merged this PR into `master`. Thanks again!

PR #383, opened 2023-03-02
Fix #200

### LINKED ISSUE #200: Argument `dis` of `family$deviance()`
The `family$deviance()` call in line https://github.com/stan-dev/projpred/blob/211e3979e7cc41dba7bdeb82e7a00e0ae26f2a07/R/glmfun.R#L67 lacks argument `dis`. On purpose? I would have thought it was needed, at least for the Gaussian family. Or do we have a unit scale here in case of the Gaussian family?

fweber144:
This tries to fix #200, but I get failing unit tests, so the fix is probably not correct. Or perhaps #200 was not even a bug in the first place.


fweber144:
In a personal communication with @jpiironen, it turned out that the missing specification of `dis` was indeed on purpose: For the Gaussian family (the only relevant case here), `dis` simply scales and shifts the deviance and thus also `loss`, so that `dis` simply scales the gradient of `loss` with respect to the regression coefficients or mu. And the only place where `loss` is used seems to be a unit test checking the approximate equivalence of analytic and finite-difference formulations of this gradient.

So I changed this PR to document this in the code.

Issue #289, opened 2022-03-17
Adapt `summary()` and `print()` output to `pct_solution_terms_cv`


fweber144:
I think the `summary()` and `print()` output for `vsel` objects from `cv_varsel()` needs to be adapted to properly account for `pct_solution_terms_cv`. I'm not sure yet about the best way to achieve this, so I'll leave this open for discussion. At the very end of this comment, I started some thoughts.

Currently, only the solution path from the full-data search is shown. However, the CV folds may have differing solution paths (this is why `pct_solution_terms_cv` exists), at least as long as we have the search included in the CV, i.e., LOO CV with `validate_search = TRUE` or K-fold CV (the latter currently only supports `validate_search = TRUE`). If `pct_solution_terms_cv` is not incorporated into the `summary()` and `print()` output (as is currently the case), users might get the false impression that there is no uncertainty with respect to the solution path and that the printed values of the performance measures (e.g., the ELPD values) are based on the full-data search (which they don't, because they are based on the cross-validated searches).

Example:
```r
options(mc.cores = parallel::detectCores(logical = FALSE))
data("df_gaussian", package = "projpred")
df_gaussian <- df_gaussian[1:41, ]
dat <- data.frame(y = df_gaussian$y, df_gaussian$x)
library(rstanarm)
rfit <- stan_glm(y ~ X1 + X2 + X3 + X4 + X5 + X6 + X7 + X8 + X9 + X10,
                 data = dat,
                 seed = 1140350788)

library(projpred)
cvvs <- cv_varsel(rfit,
                  cv_method = "kfold",
                  K = 2,
                  nclusters = 3,
                  nclusters_pred = 5,
                  seed = 963201)
```
In this case, the CV folds have different solution paths: `cvvs$pct_solution_terms_cv` gives
```
      size X1  X3  X6  X5  X9  X7  X2 X4 X10  X8
 [1,]    1  1 0.0 0.0 0.0 0.0 0.0 0.0  0   0 0.0
 [2,]    2  0 0.5 0.5 0.0 0.0 0.0 0.0  0   0 0.0
 [3,]    3  0 0.0 0.5 0.0 0.0 0.5 0.0  0   0 0.0
 [4,]    4  0 0.0 0.0 0.5 0.0 0.0 0.0  0   0 0.5
 [5,]    5  0 0.5 0.0 0.0 0.5 0.0 0.0  0   0 0.0
[...]

PR #406, opened 2023-04-19
New functions for investigating predictor rankings

fweber144:
This PR starts to address #289.

In particular, this adds the new functions `ranking()` and `props()`, making output element `pct_solution_terms_cv` of `vsel` objects obsolete (but requiring the addition of output element `solution_terms_cv` to `vsel` objects). This also adds a new `plot()` method for visualizing the output of `props()` (more precisely, two `plot()` methods are added, but `plot.vsel()` is only a shortcut for first applying `ranking.vsel()` and then `plot.ranking()`). For the user-facing changes, see the changes in `NEWS.md`.

Quick illustration of the new functions:
```r
library(projpred) # Use devtools::load_all() if installation is not desired.

dat_gauss <- data.frame(y = df_gaussian$y, df_gaussian$x)
fit <- rstanarm::stan_glm(
  y ~ X1 + X2 + X3 + X4 + X5, family = gaussian(), data = dat_gauss,
  QR = TRUE, chains = 2, iter = 1000, refresh = 0, seed = 9876
)
cvvs <- cv_varsel(fit, cv_method = "kfold", nclusters_pred = 20, seed = 5555)

# Extract predictor rankings:
rk <- ranking(cvvs)

# Compute ranking proportions:
( pr_rk <- props(rk) )
```
```
    predictor
size X1 X5 X3  X2  X4
   1  1  0  0 0.0 0.0
   2  0  1  0 0.0 0.0
   3  0  0  1 0.0 0.0
   4  0  0  0 0.6 0.4
   5  0  0  0 0.4 0.6
attr(,"class")
[1] "props"
```
```r
# Visualize the ranking proportions:
plot(pr_rk)
```
Rplot-01
```r
# For longer predictor names:
plot(pr_rk, text_angle = 45)
```
Rplot-02
```r
# Compute cumulated ranking proportions:
( pr_rk_cumul <- props(rk, cumulate = TRUE) )
```
```
     predictor
size  X1 X5 X3  X2  X4
  <=1  1  0  0 0.0 0.0
  <=2  1  1  0 0.0 0.0
  <=3  1  1  1 0.0 0.0
  <=4  1  1  1 0.6 0.4
  <=5  1  1  1 1.0 1.0
attr(,"class")
[1] "cumulprops" "props" 
```
```r
# Visualize the cumulated ranking proportions:
plot(pr_rk_cumul)
```
Rplot-03
```r
# For longer predictor names:
plot(pr_rk_cumul, text_angle = 45)
```
Rplot-04

Dummy ranking proportions for illustrating the color gradient:
```r
nterms_dummy <- 10L
pr_dummy <- matrix(
  seq(0, 1, length.out = nterms_dummy^2),
  nrow = nterms_dummy, ncol = nterms_dummy,
  dimnames = list("size" = as.character(seq_len(nterms_dummy)),
                  "predictor" = paste0("x", seq_len(nterms_dummy)))
)
class(pr_dummy) <- "props"
plot(pr_dummy)

```
Rplot

As mentioned above, this PR is only a first step towards resolving #289 completely. In the future, I will:

* Illustrate the usage of these new functions in the main vignette.
* Create a PR adding another column to the `print.vselsummary()` table, containing the diagonal of the matrix of the non-cumulated ranking proportions.
* Create a PR adding the full-data solution path on the x-axis of the ordinary `plot.vsel()` predictive performance plot, again with the diagonal of the matrix of the non-cumulated ranking proportions. This will be done with the option to restrict this solution path and the ranking proportions to the first few terms. By default, I think we should not restrict them, for consistency with  the `print.vselsummary()` table.


avehtari:
Why `props` and not `proportions`?

Otherwise, the illustration looks good

NEWS.md
```diff
@@ -4,6 +4,15 @@ If you read this from a place other than <https://mc-stan.org/projpred/news/inde
 
 # projpred 2.5.0.9000
 
+## Major changes
+
+* Until now, `solution_terms.vsel()` was the only function for extracting a solution path from `vsel` objects (returned by `varsel()` or `cv_varsel()`), namely the solution path from the full-data search. This function is unchanged by this release, to avoid breaking changes for users. However, in anticipation of a larger overhaul of the **projpred** user interface, this release introduces some new functions that are useful for accessing and investigating solution paths (which are now termed *predictor rankings* by these new functions, a term that is hopefully easier to grasp for new users):
```
avehtari, line 9: I think this could be ok, but you could ask additional comments for the term predictor ranking from the workflow group and in Stan slack

NEWS.md
```diff
@@ -4,6 +4,15 @@ If you read this from a place other than <https://mc-stan.org/projpred/news/inde
...
+
+    + Added a new function called `ranking()` which returns the predictor ranking from the full-data search and possibly also the predictor rankings from fold-wise searches in case of cross-validation (CV). (More precisely, `ranking()` is a generic. The only method is `ranking.vsel()`. The output is of class `ranking`.)
+    + Added a new function called `props()` which computes ranking proportions (across CV folds, see `?props` for details) from fold-wise predictor rankings. (More precisely, `props()` is a generic. The main method is `props.ranking()`, but as a shortcut, `props.vsel()` has also been added. The output is of class `props`.)
```
avehtari, line 12: As `prop` is a common noun, I think it would be better to not shorten and use `proportions`

avehtari (commented):
I checked the documentation parts, too

fweber144:
> Why `props` and not `proportions`?

`proportions()` was also my first idea, but it already exists in base R (https://stat.ethz.ch/R-manual/R-devel/library/base/html/proportions.html) and unfortunately, it's not a generic. Since `prop.table()` is an alias for `proportions()` in base R, I thought it would make sense to abbreviate `proportions()` by `props()`.

NEWS.md
fweber144: Yes, I'll do that. Thanks.

NEWS.md
fweber144: That's true, but see my answer [below](https://github.com/stan-dev/projpred/pull/406#issuecomment-1527009062).

NEWS.md
fweber144: I haven't received objections against "predictor ranking", so I think we can go for that term.

NEWS.md
fweber144: We now agreed to rename `props()` to `cv_proportions()` in an upcoming PR.

fweber144:
> > Why `props` and not `proportions`?
> 
> `proportions()` was also my first idea, but it already exists in base R (https://stat.ethz.ch/R-manual/R-devel/library/base/html/proportions.html) and unfortunately, it's not a generic. Since `prop.table()` is an alias for `proportions()` in base R, I thought it would make sense to abbreviate `proportions()` by `props()`.

We now agreed to rename `props()` to `cv_proportions()` in an upcoming PR.

avehtari (approved):
As discussed, ok now

PR #424, opened 2023-06-16
testing removal of except function

### LINKED ISSUE #423: Incompatability with R version <4.2.0
# Summary

The CV varsel function uses an argument which is new to R 4.2.0, see [here](https://github.com/stan-dev/projpred/blob/190a07d4e90d4d2bdc9d8b03a46fcd34a8b05e71/R/cv_varsel.R#LL964C3-L965C64). This led to an error when running this [reprex](https://github.com/l-gorman/projpred_issue_reprex).

Issue discussed in more detail on the [Stan Forum](https://discourse.mc-stan.org/t/cv-varsel-error-infinite-or-missing-values-in-x/31703/10). 

I have forked the repo [here](https://github.com/l-gorman/projpred/tree/simplify-2-array-bugfix). I will test the changes on the Reprex and see if it works. If so would you mind if I open a PR? Feel free to say no, as it is a very minor change!!!

Apologies I could not run the test suite, every time I try to run it my R session aborts!

Thanks again for all the great work with projpred, and all the support!!!

l-gorman:
# Removal of `Except` Argument in CV_Varsel Simplify2 Array

Attempt to fix #423. Very simple fix (unless I am missing something fundamental), I am removing the `except=NULL` argument, for the simplify2array function, called in the `kfold_varsel` function. I ran through with the example from the `cv_varsel` documentation.

```
dat_gauss <- data.frame(y = df_gaussian$y, df_gaussian$x)
  fit <- rstanarm::stan_glm(
    y ~ X1 + X2 + X3 + X4 + X5, family = gaussian(), data = dat_gauss,
    QR = TRUE, chains = 2, iter = 1000, refresh = 0, seed = 9876
  )
  # Run cv_varsel() (with small values for `K`, `nterms_max`, `nclusters`,
  # and `nclusters_pred`, but only for the sake of speed in this example;
  # this is not recommended in general):

  ref_model <- get_refmodel(fit)

  cvvs <- cv_varsel(object=ref_model,  
  method ="kfold",
  cv_method = if (!inherits(object, "datafit")) "LOO" else "kfold",
  ndraws = NULL,
  nclusters = 5,
  ndraws_pred = 10,
  nclusters_pred = NULL,
  refit_prj = !inherits(object, "datafit"),
  nterms_max = 3,
  penalty = NULL,
  verbose = TRUE,
  nloo = NULL,
  K =2,
  lambda_min_ratio = 1e-5,
  nlambda = 150,
  thresh = 1e-6,
  regul = 1e-4,
  validate_search = TRUE,
  seed = 5555,
  search_terms = NULL,
  parallel = getOption("projpred.prll_cv", FALSE))

```

I compared the outputs of these two calls:

```
sub_foldwise <- simplify2array(lapply(res_cv, "[[", "summaries_sub"),
                                 higher = FALSE)

sub_foldwise <- simplify2array(lapply(res_cv, "[[", "summaries_sub"),
                                 higher = FALSE, except=NULL)

```

And they appear to be the same (although I am no expert so I may be missing something. I am currently running this an an example dataset of my own to also see if this works. 

I know this is such a minor change, but I hope it helps a little! Sorry again for all of the hassle with my questions :)

### DIFF
--- a/R/cv_varsel.R
+++ b/R/cv_varsel.R
@@ -962,7 +962,7 @@ kfold_varsel <- function(refmodel, method, nterms_max, ndraws,
 
   # Handle the submodels' performance evaluation results:
   sub_foldwise <- simplify2array(lapply(res_cv, "[[", "summaries_sub"),
-                                 higher = FALSE, except = NULL)
+                                 higher = FALSE)
   sub <- apply(sub_foldwise, 1, rbind2list)
   idxs_sorted_by_fold <- unlist(lapply(list_cv, function(fold) {
     fold$omitted


fweber144:
Thank you for creating this PR. Unfortunately, `except` cannot simply be removed because it is needed for the case where a forward search is run with `nterms_max = 0` (which should not occur in practice, but still, this is a supported case, so we need to support it further). I would be inclined to set the minimum required R version to 4.2.0, but if you need to have an official projpred version which supports R < 4.2.0, then I can try to re-use the relevant code which existed before commit e11edc75e330be205a6a28b619eaeb4f11b7a2f7.

l-gorman:
I understand that we cannot include this, see full response in #423

PR #439, opened 2023-08-15
Enhance some output in case of weighted projected draws (by relying on the posterior package)

### LINKED ISSUE #206: Package `posterior` for post-processing of clustered draws
Package `posterior` offers general handling of parameter draws. I'll create a documentation PR showing how to use `posterior` together with the output from `as.matrix.projection()`. But `posterior` also has (limited) support for weighted draws (see stan-dev/posterior#184) and in particular, it offers function `posterior::resample_draws()`. Therefore, it would be nice to make the cluster weights accessible to the user.

Currently, only `proj_predict()` seems to take the clustering into account. My original idea was to offer the resampling of clustered draws at other places within `projpred`. This was already mentioned [here](https://github.com/stan-dev/projpred/pull/135#issue-650752848), but at that time not with the connection to `posterior` yet. The `posterior` package is more general and so I think it should suffice (and would be preferable) to just return the cluster weights, so they can be used in `posterior`.

fweber144:
Fixes #206. There are some larger deviations from what was suggested in #206; in particular, methods and arguments for obtaining **posterior**-formatted output are implemented directly instead of just providing the weights of the projected draws.

The behavior of `proj_predict()` remains essentially unchanged because the resampling already implemented there is more desirable than a "final-step" resampling which would typically be the type of resampling conducted with the **posterior** package: The current resampling behavior is to resample the projected draws (yielding 1000 resampled projected draws by default) and then to draw from the posterior-projection predictive distributions (PPPDs; there is one such distribution per observation) conditional on each of these resampled projected draws in turn. What I mean with "final-step" resampling is to draw from the PPPDs conditional on the non-resampled projected draws and then to resample the draws from the PPPDs. Since this would usually give a number of unique draws from the PPPDs equal to the number of non-resampled projected draws (at least for continuous families like the Gaussian family), this "final-step" resampling is less desirable.


fweber144:
Forgot to mention that `library(posterior)` now needs to be run before `devtools::document()` in order to have comments in the "Usage" documentation section for `as_draws_matrix.projection()` and `as_draws.projection()` stating that these are S3 methods. This is purely cosmetic, however. (Running `devtools::document()` without `library(posterior)` beforehand is still possible, it just causes these cosmetic comments not to appear.)

PR #467, opened 2023-10-26
Add LFO-CV to projpred

yannmclatchie:
This PR brings to `projpred`:
1. a new `ordered` search method, where we simply append the next predictor in time; and,
2. one-step-ahead LFO-CV for submodel validation.


fweber144:
Just for future reference: The following shouldn't be forgotten before merging:

- [ ] Unit tests
- [ ] `NEWS.md` entry (also mentioning you and possibly other contributors)

fweber144:
I think we should mention somewhere that within projpred, "LFO CV" actually means "1-SAP-LFO CV" (one-step-ahead prediction [or predictive?] leave-future-out cross-validation).

PR #468, opened 2023-10-27
`plot.vsel()`: Add title and subtitle

fweber144:
This adds a title and a subtitle to the plot produced by `plot.vsel()`, with the subtitle mentioning the nominal coverage as well as the type of the confidence intervals (CIs) explicitly.

The hope is that there will be less confusion about these CIs (in particular, whether these are "+/- 1 SE" or "+/- 2 SE" CIs; in general, they are neither of these two: only in case of the normal approximation, we can say that the default (due to `alpha = 2 * pnorm(-1)`) are "+/- 1 SE" CIs).

Illustration:
```r
dat <- data.frame(y = df_gaussian$y, df_gaussian$x)
library(rstanarm)
rfit <- stan_glm(y ~ X1 + X2 + X3 + X4 + X5,
                 data = dat,
                 chains = 1,
                 iter = 500,
                 seed = 1140350788,
                 refresh = 0)
devtools::load_all(".")
vs <- varsel(rfit,
             nclusters = 3,
             nclusters_pred = 5,
             nterms_max = 4,
             seed = 46782345)
plot(vs)
plot(vs, alpha = 0.05)
plot(vs, alpha = 0.05, stats = "rmse")
plot(vs, alpha = 0.05, stats = c("mlpd", "rmse"))

```
produces the following plots (in this order):

vs
vs95
vs95_rmse
vs_mlpd_rmse

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -12,6 +12,7 @@ If you read this from a place other than <https://mc-stan.org/projpred/news/inde
 ## Minor changes
 
 * In `as.matrix.projection()`, `nm_scheme = "auto"` is deprecated. Please use `nm_scheme = NULL` instead.
+* The plot produced by `plot.vsel()` now includes a title and a subtitle, with the subtitle mentioning the nominal coverage as well as the type of the confidence intervals explicitly. (GitHub: #468)
 
 ## Bug fixes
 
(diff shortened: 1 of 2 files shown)


fweber144:
To omit title and/or subtitle in the illustration above, one can use:
```r
plot(vs) + labs(title = NULL)
plot(vs) + labs(subtitle = NULL)
plot(vs) + labs(title = NULL, subtitle = NULL)
```

PR #492, opened 2023-12-28
remove QR=TRUE when prior=hs(...)

### LINKED ISSUE #491: Remove QR=TRUE from the vignette
vignettes/projpred.Rmd defines the reference model using RHS prior and `QR=TRUE`, but `QR=TRUE` should not be used in the combination with RHS or other (weakly) informative prior as stated by RStanArm doc https://mc-stan.org/rstanarm/reference/QR-argument.html
> However, because when QR is TRUE the prior argument applies to the coefficients relative to Q∗
 (and those are not very interpretable), setting QR=TRUE is only recommended if you do not have an informative prior for the regression coefficients or if the only informative prior is on the last regression coefficient (in which case you should set autoscale = FALSE when specifying such priors).

Thus `QR=TRUE` should be removed when `prior=hs(...)`

avehtari:
fixes #491

### DIFF
--- a/vignettes/latent.Rmd
+++ b/vignettes/latent.Rmd
@@ -147,7 +147,7 @@ refm_fit_poiss <- stan_glm(
   ### Only for the sake of speed (not recommended in general):
   chains = 2, iter = 1000,
   ###
-  QR = TRUE, refresh = 0
+  refresh = 0
 )
 ```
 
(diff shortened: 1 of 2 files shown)


fweber144:
Thanks, I will take a look at this asap (I think there are more places in the docs where we need to change this).

fweber144 (approved):
Should be fine. I checked for other occurrences, but there don't seem to be any which would have to be updated.

Thanks again, merging now.

PR #493, opened 2024-01-16
Clarify refit_prj argument

### LINKED ISSUE #489: Clarify doc for refit_prj
cv_varsel doc says:
> refit_prj: For the evaluation part, should the submodels along the predictor ranking be fitted again (TRUE) or should their fits from the search part be re-used (FALSE)?

I think this sentence would need to be clarified to avoid possible confusion with refitting meaning running MCMC. Maybe rewrite
as
> refit_prj: For the evaluation part, should the submodels along the predictor ranking be projected again using `ndraws_pred` draws or `nclusters_pred` clusters (TRUE) or should their projections from the search part, which used `ndraws` draws or `nclusters` clusters, be re-used (FALSE)?

avehtari:
Fixes #489

### DIFF
--- a/R/varsel.R
+++ b/R/varsel.R
@@ -18,9 +18,11 @@
 #' @param method The method for the search part. Possible options are
 #'   `"forward"` for forward search and `"L1"` for L1 search. See also section
 #'   "Details" below.
-#' @param refit_prj For the evaluation part, should the submodels along the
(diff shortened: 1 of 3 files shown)


fweber144 (approved):
Thanks, this is fine. I'll add to the [to-do list](https://github.com/fweber144/todo_projpred/) what I mentioned in https://github.com/stan-dev/projpred/issues/489#issuecomment-1887250582.

PR #496, opened 2024-04-10
subsampling LOO estimates with diff-est-srs-wor start

avehtari:
Work in progress. Tested now with family="normal" and stats %in% c("elpd", "mlpd", "gmpd", "mse", "rmse")

- in `cv_varsel` if nloo<n and fast PSIS-LOO result is not yet available, fast PSIS-LOO result is computed
- in `cv_varsel` if nloo<n, fast PSIS-LOO result is stored in slot $summaries_fast
- the subsampling indices are stored in slot $loo_inds
- the actual subsampling estimating happens in summary_funs.R get_stat()
- removed some NA checking and need to recheck if those need to put back
- improved quantiles for "mse" if the value is close to 0 (can't get negative lq)

Next
- add support for stats %in% c("acc","pctcorr","auc")
- test with other families than normal

PR #498, opened 2024-04-12
Fix pareto warnings

### LINKED ISSUE #490: Don't warn about slightly large khats
See https://github.com/stan-dev/loo/issues/234. If the default `refit_prj=TRUE` and `ndraws_pred=400` are used, then the khat threshold would be 0.62

avehtari:
Fixes #490

### DIFF
--- a/R/cv_varsel.R
+++ b/R/cv_varsel.R
@@ -785,9 +785,10 @@ loo_varsel <- function(refmodel, method, nterms_max, ndraws,
         if (no_psis_eval) {
           if (getOption("projpred.warn_psis", TRUE)) {
             warning(
-              "Using standard importance sampling (SIS), as the number of draws",
-              "or clusters is too small for PSIS. For improved accuracy",
-              "increase the number of draws or clusters, or use K-fold-CV."
+              "Using standard importance sampling (SIS), as the number of ",
+              "draws or clusters is too small for PSIS. For improved ",
+              "accuracy, increase the number of draws or clusters, or use ",
+              "K-fold-CV."
             )
           }
           # Use loo::sis().
@@ -840,7 +841,8 @@ loo_varsel <- function(refmodel, method, nterms_max, ndraws,
           n07 = sum(pareto_k_eval > .ps_khat_threshold(dim(psisloo)[1])), n = n,
           khat_threshold = .ps_khat_threshold(dim(sub_psisloo)[1]),
           warn_txt = paste0(
-            "Some (%d / % d) Pareto k's for the reference model's PSIS-LOO weights given ",
+            "Some (%d / % d) Pareto k's for the reference model's PSIS-LOO ",
+            "weights given ",
             ifelse(clust_used_eval,
                    paste0(nclusters_pred, " clustered "),
                    paste0(ndraws_pred, " posterior ")),
--- a/R/cv_varsel.R
+++ b/R/cv_varsel.R
@@ -613,7 +613,7 @@ loo_varsel <- function(refmodel, method, nterms_max, ndraws,
   # argument, another reason is that they can quickly become as costly as
   # K-fold-CV).
   warn_pareto(n07 = sum(pareto_k > 0.7), n = n,
-              khat_threshold = loo:::ps_khat_threshold(dim(psisloo)[1]),
+              khat_threshold = .ps_khat_threshold(dim(psisloo)[1]),
     warn_txt = paste0("Some Pareto k's for the reference model's ",
       "PSIS-LOO weights are >%s (%d / %d).",
       "\n\nMoment matching (see the `loo` package), mixture importance ",
@@ -642,10 +642,10 @@ loo_varsel <- function(refmodel, method, nterms_max, ndraws,
       log_ratios = -loglik_forPSIS
     )
     # The k-values are h-specific (expectation-specific) here (see Vehtari et
-    # al., 2024, <https://doi.org/10.48550/arXiv.1507.02646>, beginning of
+    # al., 2024, <https://jmlr.org/papers/v25/19-556.html>, beginning of
     # section 3, section 3.2.8, appendix D, and appendix E).
     warn_pareto(n07 = sum(y_lat_E$pareto_k > 0.7), n = n,
-                khat_threshold = loo:::ps_khat_threshold(dim(psisloo)[1]),
+                khat_threshold = .ps_khat_threshold(dim(psisloo)[1]),
       warn_txt = paste0("In the recalculation of the latent response values, ",
         "some expectation-specific Pareto k-values are >%s (%d / % d). ",
        "In general, we recommend K-fold-CV in this case."
@@ -787,8 +787,8 @@ loo_varsel <- function(refmodel, method, nterms_max, ndraws,
           # Use loo::psis().
           # Usually, we have a small number of projected draws here (400 by
           # default), which means that the 'loo' package will automatically
-          # perform the regularization from Vehtari et al. (2022,
-          # <https://doi.org/10.48550/arXiv.1507.02646>, appendix G).
+          # perform the regularization from Vehtari et al. (2024,
+          # <https://jmlr.org/papers/v25/19-556.html>, appendix G).
           importance_sampling_nm <- "psis"
         }
       } else {
@@ -822,7 +822,7 @@ loo_varsel <- function(refmodel, method, nterms_max, ndraws,
       if (importance_sampling_nm == "psis") {
         pareto_k_eval <- loo::pareto_k_values(sub_psisloo)
         warn_pareto(n07 = sum(pareto_k_eval > 0.7), n = n,
-                    khat_threshold = loo:::ps_khat_threshold(dim(sub_psisloo)[1]),
+                    khat_threshold = .ps_khat_threshold(dim(sub_psisloo)[1]),
           warn_txt = paste0(
             "Some Pareto k's for the reference model's ",
             "PSIS-LOO weights for ",
@@ -1625,4 +1625,24 @@ loo_subsample_pps <- function(nloo, lppd) {
   wcv <- wcv / sum(wcv)
 
   return(nlist(inds, wcv))
+}
+
+#' Pareto-smoothing k-hat threshold
+#'
+#' Copied from loo package. Remove after loo package exposes this.
+#'
+#' Given sample size S computes khat threshold for reliable Pareto
+#' smoothed estimate (to have small probability of large error). See
+#' section 3.2.4, equation (13). Sample sizes 100, 320, 1000, 2200,
+#' 10000 correspond to thresholds 0.5, 0.6, 0.67, 0.7, 0.75. Although
+#' with bigger sample size S we can achieve estimates with small
+#' probability of large error, it is difficult to get accurate MCSE
+#' estimates as the bias starts to dominate when k > 0.7 (see Section 3.2.3).
+#' Thus the sample size dependend k-ht threshold is capped at 0.7.
+#' @param S sample size
+#' @param ... unused
+#' @return threshold
+#' @noRd
+.ps_khat_threshold <- function(S, ...) {
+  min(1 - 1 / log10(S), 0.7)
 }
--- a/R/cv_varsel.R
+++ b/R/cv_varsel.R
@@ -76,13 +76,6 @@
 #'   used to guide further actions and the optimistic bias can be greatly
 #'   reduced by using `validate_search=TRUE`.
 #'
-#'   For PSIS-LOO-CV, \pkg{projpred} calls [loo::psis()] (or, exceptionally,
-#'   [loo::sis()], see below) with `r_eff = NA`. This is only a problem if there
-#'   was extreme autocorrelation between the MCMC iterations when the reference
-#'   model was built. In those cases however, the reference model should not
-#'   have been used anyway, so we don't expect \pkg{projpred}'s `r_eff = NA` to
-#'   be a problem.
-#'
 #'   PSIS uses Pareto-\eqn{\hat{k}} diagnostic to assess the reliability of PSIS-LOO-CV.
 #'   See [loo::loo-glossary] for how to interpret the Pareto-\eqn{\hat{k}} values and
 #'   the warning thresholds. \pkg{projpred} does not support the usually recommended
@@ -97,6 +90,13 @@
 #'   when using moment-matching, mixture importance sampling, or `reloo`-ing, we
 #'   recommend to use $K$-fold-CV within `projpred`.
 #' 
+#'   For PSIS-LOO-CV, \pkg{projpred} calls [loo::psis()] (or, exceptionally,
+#'   [loo::sis()], see below) with `r_eff = NA`. This is only a problem if there
+#'   was extreme autocorrelation between the MCMC iterations when the reference
+#'   model was built. In those cases however, the reference model should not
+#'   have been used anyway, so we don't expect \pkg{projpred}'s `r_eff = NA` to
+#'   be a problem.
+#'
 #'   PSIS cannot be used if the number of draws or clusters is too small. In such
 #'   cases, \pkg{projpred} resorts to standard importance sampling (SIS) and
 #'   shows a message about this. Throughout the documentation, the term "PSIS" is
--- a/R/cv_varsel.R
+++ b/R/cv_varsel.R
@@ -666,7 +666,7 @@ loo_varsel <- function(refmodel, method, nterms_max, ndraws,
       n07 = sum(y_lat_E$pareto_k > .ps_khat_threshold(dim(psisloo)[1])), n = n,
       khat_threshold = .ps_khat_threshold(dim(psisloo)[1]),
       warn_txt = paste0(
-        "In the recalculation of the latent response values, some (%d / % d) ",
+        "In the recalculation of the latent response values, some (%d / %d) ",
         "expectation-specific Pareto k-values are > %s.\n",
         "In general, we recommend K-fold-CV in this case."
       )
@@ -841,7 +841,7 @@ loo_varsel <- function(refmodel, method, nterms_max, ndraws,
           n07 = sum(pareto_k_eval > .ps_khat_threshold(dim(psisloo)[1])), n = n,
           khat_threshold = .ps_khat_threshold(dim(sub_psisloo)[1]),
           warn_txt = paste0(
-            "Some (%d / % d) Pareto k's for the reference model's PSIS-LOO ",
+            "Some (%d / %d) Pareto k's for the reference model's PSIS-LOO ",
             "weights given ",
             ifelse(clust_used_eval,
                    paste0(nclusters_pred, " clustered "),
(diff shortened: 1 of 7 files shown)


R/cv_varsel.R
```diff
@@ -609,26 +609,24 @@ loo_varsel <- function(refmodel, method, nterms_max, ndraws,
...
-    n07 = sum(pareto_k > 0.7), n05 = sum(0.7 >= pareto_k & pareto_k > 0.5),
-    warn_txt_start = paste0("In the calculation of the reference model's ",
-                            "PSIS-LOO CV weights, "),
-    warn_txt_mid_common = paste0(" (out of ", n, ") Pareto k-values are "),
-    warn_txt_end = paste0(
-      ". Moment matching (see the loo package), mixture importance ",
-      "sampling (see the loo package), and `reloo`-ing (see the brms package) ",
+  # K-fold-CV).
+  warn_pareto(n07 = sum(pareto_k > 0.7), n = n,
+              khat_threshold = loo:::ps_khat_threshold(dim(psisloo)[1]),
```
fweber144, line 616: Doesn't `:::` cause `R CMD check` to complain?

R/cv_varsel.R
```diff
@@ -644,18 +642,15 @@ loo_varsel <- function(refmodel, method, nterms_max, ndraws,
       log_ratios = -loglik_forPSIS
     )
     # The k-values are h-specific (expectation-specific) here (see Vehtari et
-    # al., 2022, <https://doi.org/10.48550/arXiv.1507.02646>, beginning of
+    # al., 2024, <https://doi.org/10.48550/arXiv.1507.02646>, beginning of
```
fweber144, line 645: Yey! :) Just noting that the DOI needs to be updated and perhaps this reference occurs at other places as well where it needs to be updated, too.

R/cv_varsel.R
```diff
@@ -644,18 +642,15 @@ loo_varsel <- function(refmodel, method, nterms_max, ndraws,
...
-      n05 = sum(0.7 >= y_lat_E$pareto_k & y_lat_E$pareto_k > 0.5),
-      warn_txt_start = paste0("In the recalculation of the latent response ",
-                              "values, "),
-      warn_txt_mid_common = paste0(
-        " (out of ", n, ") expectation-specific Pareto k-values are "
-      ),
-      warn_txt_end = ". In general, we recommend K-fold CV in this case."
-    )
+    warn_pareto(n07 = sum(y_lat_E$pareto_k > 0.7), n = n,
+                khat_threshold = loo:::ps_khat_threshold(dim(psisloo)[1]),
```
fweber144, line 648: For `:::`, see above.

R/cv_varsel.R
```diff
@@ -823,24 +818,22 @@ loo_varsel <- function(refmodel, method, nterms_max, ndraws,
...
-            "thinned posterior draws), "
-          ),
-          warn_txt_mid_common = paste0(
-            " (out of ", nloo, ") Pareto k-values are "
-          ),
-          warn_txt_end = paste0(
-            ". Watch out for warnings thrown by the original-draws Pareto ",
-            "smoothing to see whether it makes sense to increase the number ",
+        warn_pareto(n07 = sum(pareto_k_eval > 0.7), n = n,
+                    khat_threshold = loo:::ps_khat_threshold(dim(sub_psisloo)[1]),
```
fweber144, line 825: For `:::`, see above.

R/cv_varsel.R
avehtari: fixed

R/cv_varsel.R
avehtari: fixed

avehtari:
fixed ::: and ref, `devtools::check()` reported
```
❯ checking compiled code ... OK
   WARNING
  ‘qpdf’ is needed for checks on size reduction of PDFs

❯ checking package dependencies ... NOTE
  Packages suggested but not available for checking:
    'optimx', 'unix', 'glmnet', 'future.callr'

❯ checking Rd cross-references ... NOTE
  Package unavailable to check Rd xrefs: ‘unix’

❯ checking compilation flags used ... NOTE
  Compilation used the following non-portable flag(s):
    ‘-march=native’

0 errors ✔ | 1 warning ✖ | 3 notes ✖
```

R/cv_varsel.R
```diff
@@ -609,26 +609,24 @@ loo_varsel <- function(refmodel, method, nterms_max, ndraws,
...
-                            "PSIS-LOO CV weights, "),
-    warn_txt_mid_common = paste0(" (out of ", n, ") Pareto k-values are "),
-    warn_txt_end = paste0(
-      ". Moment matching (see the loo package), mixture importance ",
-      "sampling (see the loo package), and `reloo`-ing (see the brms package) ",
+  # K-fold-CV).
+  warn_pareto(n07 = sum(pareto_k > 0.7), n = n,
+              khat_threshold = .ps_khat_threshold(dim(psisloo)[1]),
+    warn_txt = paste0("Some Pareto k's for the reference model's ",
+      "PSIS-LOO weights are >%s (%d / %d).",
```
fweber144, line 618: I would prefer moving ` (%d / %d)` after `"Some` at the beginning of the message because otherwise, unexperienced users might think the ratio relates to the `khat_threshold` printed by `%s`.

R/cv_varsel.R
```diff
@@ -609,26 +609,24 @@ loo_varsel <- function(refmodel, method, nterms_max, ndraws,
...
-  warn_pareto(
-    n07 = sum(pareto_k > 0.7), n05 = sum(0.7 >= pareto_k & pareto_k > 0.5),
-    warn_txt_start = paste0("In the calculation of the reference model's ",
-                            "PSIS-LOO CV weights, "),
-    warn_txt_mid_common = paste0(" (out of ", n, ") Pareto k-values are "),
-    warn_txt_end = paste0(
-      ". Moment matching (see the loo package), mixture importance ",
-      "sampling (see the loo package), and `reloo`-ing (see the brms package) ",
+  # K-fold-CV).
+  warn_pareto(n07 = sum(pareto_k > 0.7), n = n,
```
fweber144, line 615: If I read this (and the definition of `warn_pareto()` correctly, then `n07` still refers to the fixed threshold of 0.7, but the `khat_threshold` printed by `%s` refers to the updated threshold. Shouldn't they be consistent? So `n07` would need to be updated as well?

R/cv_varsel.R
fweber144: At that occasion, do we perhaps also want to add a short explanation where the value printed by `%s` comes from? Often, 0.7 should be printed (which is probably established enough that it gets quickly recognized by many users), but given that the recommendation of a Monte-Carlo-sample-size-specific k-hat threshold is quite recent, many users might not be aware where a value other than 0.7 (depending on the Monte Carlo sample size) comes from. I'm thinking of adding something like ` (the k-hat threshold for the given Monte Carlo sample size)` after `%s` (this would be possible if we moved the ` (%d / %d)` parenthesis up front, as suggested in my previous comment).

R/cv_varsel.R
```diff
@@ -609,26 +609,24 @@ loo_varsel <- function(refmodel, method, nterms_max, ndraws,
...
+              khat_threshold = .ps_khat_threshold(dim(psisloo)[1]),
+    warn_txt = paste0("Some Pareto k's for the reference model's ",
+      "PSIS-LOO weights are >%s (%d / %d).",
+      "\n\nMoment matching (see the `loo` package), mixture importance ",
```
fweber144, line 619: I understand that you want to separate the first part of the warning message (a description of what has been encountered) from the second part (a recommendation what to do next) and this is definitely a good idea, but it now gives warnings like
```
Warning: Some Pareto k's for the reference model's PSIS-LOO weights are >0.58 (1 / 29).

Moment matching (see the `loo` package), mixture importance sampling (see the loo package), and `reloo`-ing (see the `brms` package) are not supported by projpred. If these techniques (run outside of projpred, i.e., for the reference model only; note that `reloo`-ing may be computationally costly) result in a markedly different reference model ELPD estimate than ordinary PSIS-LOO-CV does, we recommend to use K-fold-CV within projpred.
```
Personally, I think the two parts are now too far away from each other, especially given that they will typically occur amidst a bunch of other console output. Perhaps a single `\n` is enough? Or revert to not using `\n` at all?

fweber144:
Commit 930ad81 is purely cosmetic. Feel free to revert it if you don't like it.

R/cv_varsel.R
avehtari: I would prefer much shorter messages and include the longer explanations in the docs. Maybe we should add something like LOO-glossary in `loo` https://mc-stan.org/loo/reference/loo-glossary.html ?

R/cv_varsel.R
avehtari: I guess this is a mistake, I'll check

R/cv_varsel.R
avehtari: I moved `(%d / %d)`. I made the warning message even shorter and improved the `cv_varsel()` doc.

R/cv_varsel.R
avehtari: Fixed

R/cv_varsel.R
avehtari: I made the warning shorter and improved `cv_varsel()` doc

avehtari:
I'm getting these errors when running tests
```
Failure (]8;line = 225:col = 7;file:///u/77/ave/unix/proj/projpred/tests/testthat/test_as_matrix.Rtest_as_matrix.R:225:7]8;;): as.matrix.projection() works
Snapshot of code has changed:
    old                                      | new                                         
[5] Code                                     | Code                                     [5]
[6]   print(rlang::hash(m))                  |   print(rlang::hash(m))                  [6]
[7] Output                                   | Output                                   [7]
[8]   [1] "ab77d9f8eb5a3fae48c7e63b2ee52d5a" -   [1] "932c98c86dced4ef17094f8a70b35a04" [8]

* Run `testthat::snapshot_accept('as_matrix')` to accept the change.
* Run `testthat::snapshot_review('as_matrix')` to interactively review the change.
```

I don't understand what has changed. Any ideas?

fweber144:
> I'm getting these errors when running tests
> 
> ```
> Failure (�]8;line = 225:col = 7;file:///u/77/ave/unix/proj/projpred/tests/testthat/test_as_matrix.R�test_as_matrix.R:225:7�]8;;�): as.matrix.projection() works
> Snapshot of code has changed:
>     old                                      | new                                         
> [5] Code                                     | Code                                     [5]
> [6]   print(rlang::hash(m))                  |   print(rlang::hash(m))                  [6]
> [7] Output                                   | Output                                   [7]
> [8]   [1] "ab77d9f8eb5a3fae48c7e63b2ee52d5a" -   [1] "932c98c86dced4ef17094f8a70b35a04" [8]
> 
> * Run `testthat::snapshot_accept('as_matrix')` to accept the change.
> * Run `testthat::snapshot_review('as_matrix')` to interactively review the change.
> ```
> 
> I don't understand what has changed. Any ideas?

Have you updated to R 4.4.0 and/or relevant packages like the Matrix package? I did so myself and this gave me updated snapshots, too.

R/cv_varsel.R
fweber144: Sounds good :)

R/cv_varsel.R
fweber144: Thanks

R/cv_varsel.R
fweber144: Thanks!

R/cv_varsel.R
```diff
@@ -68,22 +68,45 @@
...
+#'   See [loo::loo-glossary] for how to interpret the Pareto-\eqn{\hat{k}} values and
+#'   the warning thresholds. \pkg{projpred} does not support the usually recommended
+#'   moment-matching (see [loo::loo_moment_match()] and [brms::loo_moment_match()]),
+#'   mixture importance sampling (`vignette("loo2-mixis", package="loo")`),
+#'   or `reloo`-ing ([brms::reloo()]). If the reference model PSIS-LOO-CV
+#'   Pareto-\eqn{\hat{k}} values are good, but there are high Pareto-\eqn{\hat{k}}
+#'   values for the projected models, you can try increasing the number of draws used
+#'   for the PSIS-LOO-CV (`ndraws_pred` with  `refit_prj=TRUE`). If increasing the
+#'   number of draws does not help and if the reference model PSIS-LOO-CV
+#'   Pareto-\eqn{\hat{k}} values are high, and the PSIS-LOO-CV results change substantially
```
fweber144, line 91: Here, I'm wondering whether we want to replace "the PSIS-LOO-CV results" by some like "the reference model's PSIS-LOO-CV results" (just for clarity).

R/cv_varsel.R
```diff
@@ -68,22 +68,45 @@
...
+#'   recommend to use $K$-fold-CV within `projpred`.
+#' 
+#'   For PSIS-LOO-CV, \pkg{projpred} calls [loo::psis()] (or, exceptionally,
 #'   [loo::sis()], see below) with `r_eff = NA`. This is only a problem if there
 #'   was extreme autocorrelation between the MCMC iterations when the reference
 #'   model was built. In those cases however, the reference model should not
 #'   have been used anyway, so we don't expect \pkg{projpred}'s `r_eff = NA` to
 #'   be a problem.
 #'
-#'   PSIS cannot be used if the draws have different (i.e., nonconstant) weights
```
fweber144, line 80: This part about nonconstant weights does not appear in the new text anymore. Was this on purpose? As far as I know, `loo::psis()` is not able to deal with nonconstant weights out-of-the-box.

R/cv_varsel.R
avehtari: updated

R/cv_varsel.R
avehtari: If the draws have weights, we can just multiply them with importance weights and use Pareto diagnostics. This approach was used, e.g., in http://proceedings.mlr.press/v97/magnusson19a.html. I made the change in the code in that other PR, but forgot that the doc changes are also there.

fweber144: I see. So with PR #496, PSIS for draws with nonconstant weights will be supported. Then I'll resolve this conversation here.

fweber144:
In commit b19dfcd7db6f77b8bf53f63902492e70903662b2, you switched to roxygen2 version 7.3.1, which now gives me
```
✖ formula.R:808: S3 method `formula.gamm4` needs @export or @exportS3method tag.
✖ divergence_minimizers.R:1216: S3 method `predict.gamm4` needs @export or @exportS3method tag.
✖ divergence_minimizers.R:1179: S3 method `predict.subfit` needs @export or @exportS3method tag.
```
when running `devtools::document()`. I'll open a new issue for this.

avehtari:
- I have R 4.3.3, but it seems updating some other package updated Matrix to 1.6-5 (2024-01-06) (which happens to be the last release for pre 4.4, and 4.4 requires Matrix 1.7+)
- I didn't intentionally switch roxygen2 version

fweber144:
> * I have R 4.3.3, but it seems updating some other package updated Matrix to 1.6-5 (2024-01-06) (which happens to be the last release for pre 4.4, and 4.4 requires Matrix 1.7+)

Ok, then the snapshot changes are probably due to the Matrix package update. I'll check this on my machine asap.

>     * I didn't intentionally switch roxygen2 version

No problem, I am definitely in favor of using the most recent roxygen2 version. And I think it shouldn't be too much work to get rid of those warnings (see #501).

R/cv_varsel.R
```diff
@@ -770,14 +782,9 @@ loo_varsel <- function(refmodel, method, nterms_max, ndraws,
...
-              "In the recalculation of the reference model's PSIS-LOO CV ",
-              "weights for the performance evaluation, the number of draws ",
-              "after clustering or thinning is too small for Pareto ",
-              "smoothing. Using standard importance sampling (SIS) instead. ",
-              "Watch out for warnings thrown by the original-draws Pareto ",
-              "smoothing to see whether it makes sense to increase the number ",
-              "of draws (resulting from the clustering or thinning for the ",
-              "performance evaluation). Alternatively, K-fold CV can be used."
+              "Using standard importance sampling (SIS), as the number of draws",
+              "or clusters is too small for PSIS. For improved accuracy reliability",
```
fweber144, line 786: Is "accuracy reliability" supposed to read only "accuracy" (or "reliability")?

R/cv_varsel.R
```diff
@@ -823,24 +830,18 @@ loo_varsel <- function(refmodel, method, nterms_max, ndraws,
...
-            "of draws (resulting from the clustering or thinning for the ",
-            "performance evaluation). Alternatively, K-fold CV can be used."
+          n07 = sum(pareto_k_eval > .ps_khat_threshold(dim(psisloo)[1])), n = n,
+          khat_threshold = .ps_khat_threshold(dim(sub_psisloo)[1]),
+          warn_txt = paste0(
+            "Some (%d / % d) Pareto k's for the reference model's PSIS-LOO weights for ",
+            ifelse(clust_used_eval,
+                   paste0(nclusters_pred, " clustered "),
+                   paste0(ndraws_pred, " posterior ")),
+            "draws are > %s."
```
fweber144, line 844: This now generates warnings such as
```
Warning: Some (7 /  100) Pareto k's for the reference model's PSIS-LOO weights for 30 posterior draws are > 0.32.
```
in case of thinning. I'm not sure if this will be easy to understand for users (because this lacks the reference to thinning now and also because the double occurrence of "for" makes it slightly hard to read), but I'm fine with keeping this if you prefer so.

fweber144:
> > * I have R 4.3.3, but it seems updating some other package updated Matrix to 1.6-5 (2024-01-06) (which happens to be the last release for pre 4.4, and 4.4 requires Matrix 1.7+)
> 
> Ok, then the snapshot changes are probably due to the Matrix package update. I'll check this on my machine asap.

As far as I can tell, any snapshot changes should be due to package updates.

fweber144:
I have added a `NEWS` entry in commit 87331d6c57028b488ed999d57141ea41aba461d1. Please check if it is ok for you. Afterwards (keeping in mind my final comments above: https://github.com/stan-dev/projpred/pull/498#discussion_r1631670167 and https://github.com/stan-dev/projpred/pull/498#discussion_r1631677716), I think this PR is ready to be merged (I'll run `R CMD check` once again, though).

R/cv_varsel.R
avehtari: fixed

R/cv_varsel.R
avehtari: fixed

avehtari:
I fixed the two sentences you mentioned

R/cv_varsel.R
fweber144: thanks

R/cv_varsel.R
fweber144: thanks

fweber144:
`R CMD check` succeeded. Merging now.

PR #508, opened 2025-04-05
Add mixed deltas plot support

avehtari:
For some measures like R^2, it is more natural to look at the performance plot in the original scale even when using intervals for the difference from the reference model. With this PR, `deltas` argument accepts values `TRUE`, `FALSE` and `"mixed"`. `TRUE` and `FALSE` behave as before, and `"mixed"` uses the diff_se's but original scales for selected measures.


avehtari:
I have checked that "mixed" argument works as desired.

fweber144:
Thanks, I'll take a look.

PR #532, opened 2025-08-22
Fix #531

### LINKED ISSUE #531: 3-way interactions
I see #190 fixed the ordering problem with 2-way interactions, but the issue still persists for 3-way interactions.

``` r
library(brms)
#> Loading required package: Rcpp
#> Loading 'brms' package (version 2.22.0). Useful instructions
#> can be found by typing help('brms'). A more detailed introduction
#> to the package is available through vignette('brms_overview').
#> 
#> Attaching package: 'brms'
#> The following object is masked from 'package:stats':
#> 
#>     ar
library(projpred)
#> This is projpred version 2.9.0.
#> 
#> Attaching package: 'projpred'
#> The following object is masked from 'package:brms':
#> 
#>     do_call
nobsv <- 29
nvars <- 3
set.seed(1235)
x_mat <- matrix(rnorm(nobsv * nvars), nrow = nobsv, ncol = nvars)
x_mat <- cbind(x_mat, x_mat[, 1] * x_mat[, 2],  x_mat[, 1] * x_mat[, 3],  x_mat[, 2] * x_mat[, 3],  x_mat[, 1] * x_mat[, 2] * x_mat[, 3])
b <- c(-3, 1, 2, 3, -1, -2, 4)
dat <- data.frame(y = rnorm(nobsv, mean = x_mat %*% b, sd = 1.5), x_mat)
fitobj <- brm(y ~ X1*X2*X3,
              data = dat,
              seed = 11403,
              silent = 2,
              chains = 4, 
              cores = 4,
              refresh = 0)

vs <- varsel(fitobj, nclusters = 2, nclusters_pred = 3, seed = 457634)
ranking(vs)$fulldata
#> [1] "X1"       "X2"       "X1:X2"    "X3"       "X1:X2:X3" "X2:X3"    "X1:X3"
```

<sup>Created on 2025-08-16 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

fweber144:
Fixes #531. The problem was in `split_interaction_term()`: Before, only main-effect terms were added. Now all lower-order terms (including main-effect terms) are added.

At first, I thought it would make sense to add only lower-order terms which also occur in the full formula, but then I realized that analogously to the intercept which is always included in the submodels (https://github.com/stan-dev/projpred/issues/96#issuecomment-1018361615), all lower-order terms should be added, even those which do not occur in the full formula. The reason is that even if the true coefficient for some lower-order term in the reference model were zero, this would only hold for the reference model and would not have to hold for a submodel.

As explained in the message of commit c11bac4145ea4700bf54e774463834f6f90ed76d, argument `return_group_terms` is removed here because it was unused (always at its default of `TRUE`) and because it was easier to deal without separate `return_group_terms` cases.

Apart from #531, there might be another issue (an inconsistency in how `split_group_term()` behaves in case of `add_lower_terms = FALSE`), but I'll open a new issue for that.

Again, I requested a review from @avehtari because I could not request it from @aloctavodia. Tell me if I should request it from someone else.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,13 @@ If you read this from a place other than <https://mc-stan.org/projpred/news/inde
 
 # projpred 2.9.0.9000
 
+## Major changes
+
+* Fixed a major bug in forward search with 3-way or higher-order interactions. See "Bug fixes" below for details.
+
+## Bug fixes
+
+* Fixed a bug that caused forward search not to place lower-order terms of a 3-way or higher-order interaction term before that interaction term. (GitHub: #531, #532)
 
 # projpred 2.9.0
 
--- a/R/formula.R
+++ b/R/formula.R
@@ -286,9 +286,11 @@ flatten_group_terms <- function(terms_) {
 ## @param formula A formula for a valid model.
 ## @param return_group_terms If TRUE, return group terms as well. Default TRUE.
 ## @param data The reference model data.
+## @param add_lower_terms Passed to argument `add_lower_terms` of
+##   split_group_term() and split_interaction_term().
 ## @return a vector of all the minimal valid terms that make up for submodels.
 split_formula <- function(formula, return_group_terms = TRUE, data = NULL,
-                          add_main_effects = TRUE) {
+                          add_lower_terms = TRUE) {
   terms_ <- extract_terms_response(formula)
   group_terms <- terms_$group_terms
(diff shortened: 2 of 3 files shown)


fweber144:
> Maybe add a small test or update an existing one.

Yes, done in commit 64e456252eca59ddbe11dc71b865883e7e41d74f.

PR #534, opened 2025-08-23
Fix #533

### LINKED ISSUE #533: Inconsistency in `split_group_term()` for `add_lower_terms = FALSE`
There is an inconsistency in how `split_group_term()` behaves in case of `add_lower_terms = FALSE` (before #532, `add_lower_terms` was called `add_main_effects`, but I'll refer to the state of #532 here). This inconsistency occurs between a group-level term that has a group-level intercept and a group-level term that does *not* have a group-level intercept. This is relevant only for `project()` because only there, we have `add_lower_terms = FALSE`.

Reprex:
``` r
devtools::load_all("<path_to_projpred>")
#> ℹ Loading projpred
#> This is projpred version 2.9.0.9000.
split_formula(y ~ (x + z + x:z | g), add_lower_terms = FALSE)
#> [1] "1"         "(1 | g)"   "(x | g)"   "(z | g)"   "(x:z | g)"
split_formula(y ~ (0 + x + z + x:z | g), add_lower_terms = FALSE)
#> [1] "1"                   "x + (0 + x | g)"     "z + (0 + z | g)"
#> [4] "x:z + (0 + x:z | g)"
```

<sup>Created on 2025-08-23 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

I would have expected the same output for the first `split_formula()` call, but different output for the second `split_formula()` call (namely, without the population-level terms `x`, `z`, and `x:z` added):
``` r
devtools::load_all("<path_to_projpred>")
#> ℹ Loading projpred
#> This is projpred version 2.9.0.9000.
split_formula(y ~ (x + z + x:z | g), add_lower_terms = FALSE)
#> [1] "1"         "(1 | g)"   "(x | g)"   "(z | g)"   "(x:z | g)"
split_formula(y ~ (0 + x + z + x:z | g), add_lower_terms = FALSE)
#> [1] "1"                   "(0 + x | g)"     "(0 + z | g)"
#> [4] "(0 + x:z | g)"
```

The reason I expect it this way (and not in the way that the first `split_formula()` call should add the population-level terms `x`, `z`, and `x:z`) is the way `split_group_terms()` behaves in case of `group_intercept == TRUE` (object `group_intercept` is created inside of `split_group_terms()`):
``` r
devtools::load_all("<path_to_projpred>")
#> ℹ Loading projpred
#> This is projpred version 2.9.0.9000.
split_formula(y ~ (x + z + x:z | g))
#> [1] "1"                               "(1 | g)"                        
#> [3] "x + (x | g)"                     "z + (z | g)"                    
#> [5] "x + z + x:z + (x + z + x:z | g)" "x + (1 | g)"                    
#> [7] "z + (1 | g)"                     "x + z + x:z + (1 | g)"
split_formula(y ~ (x + z + x:z | g), add_lower_terms = FALSE)
#> [1] "1"         "(1 | g)"   "(x | g)"   "(z | g)"   "(x:z | g)"
```

<sup>Created on 2025-08-23 with [reprex v2.1.1](https://reprex.tidyverse.org)</sup>

In lines https://github.com/fweber144/projpred/blob/c11bac4145ea4700bf54e774463834f6f90ed76d/R/formula.R#L430-L492, one can see the different `add_lower_terms` cases in the `group_intercept == TRUE` case. Different `add_lower_terms` cases are missing in the `group_intercept == FALSE` case (lines https://github.com/fweber144/projpred/blob/c11bac4145ea4700bf54e774463834f6f90ed76d/R/formula.R#L493-L508), so I think it was simply forgotten to have different `add_lower_terms` cases there. I'll add a PR with a fix how I would imagine it.

fweber144:
Fixes #533 (how I would imagine it). This builds on top of branch `fix531` (from #532), so #532 should be merged first.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -11,6 +11,7 @@ If you read this from a place other than <https://mc-stan.org/projpred/news/inde
 ## Bug fixes
 
 * Fixed a bug that caused forward search not to place lower-order terms of a 3-way or higher-order interaction term before that interaction term. (GitHub: #531, #532)
+* Fixed a bug that caused `project()` to construct an incorrect internal table of predictor terms (from which argument `predictor_terms` can select one or several terms from) in case of a group-level term that does not contain a group-level intercept. (GitHub: #533, #534)
 
 # projpred 2.9.0
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -4,6 +4,13 @@ If you read this from a place other than <https://mc-stan.org/projpred/news/inde
 
 # projpred 2.9.0.9000
 
+## Major changes
+
+* Fixed a major bug in forward search with 3-way or higher-order interactions. See "Bug fixes" below for details.
+
+## Bug fixes
+
+* Fixed a bug that caused forward search not to place lower-order terms of a 3-way or higher-order interaction term before that interaction term. (GitHub: #531, #532)
 
 # projpred 2.9.0
 
--- a/R/formula.R
+++ b/R/formula.R
@@ -491,20 +491,35 @@ split_group_term <- function(term, add_lower_terms = TRUE) {
       )
     }
   } else {
-    group_terms <- lapply(lin_v, function(v) {
-      paste0(v, " + ", "(0 + ", v, " | ", group, ")")
-    })
-    group_terms <- c(group_terms, lapply(int_v, function(v) {
-      paste0(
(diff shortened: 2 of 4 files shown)


aloctavodia:
I think combining is fine in this case.

PR #542, opened 2025-11-26
suggest_size(): Return nterms_max instead of NA when no model satisfies the criterion

### LINKED ISSUE #540: suggest_size() should return full model size or nterms_max instead of NA
Currently `suggest_size()` returns NA if none of the model performances along the search path match the given criterion. It would be better to return the full model size or nterms_max instead.

aloctavodia:
Closes #540

### DIFF
--- a/R/methods.R
+++ b/R/methods.R
@@ -1820,7 +1820,7 @@ suggest_size.vsel <- function(
     if (object$nterms_max != object$nterms_all && warnings) {
       warning("Could not suggest submodel size within the searched range. ",
               "Investigate plot.vsel() to identify if the search was ",
-              "terminated too early or the criterion is too strict."
+              "terminated too early or the criterion is too strict.",
               "If this is the case, run variable selection with ",
               "larger value for `nterms_max` or use less strict criterion.")
     }


R/methods.R
```diff
@@ -1816,16 +1816,12 @@ suggest_size.vsel <- function(
...
-                "identify if the search was terminated too early. If this is ",
-                "the case, run variable selection with larger value for ",
-                "`nterms_max`.")
-      }
+    suggested_size <- object$nterms_max
+    if (object$nterms_max != object$nterms_all && warnings) {
+      warning("Could not suggest submodel size within the searched range. ",
+              "Investigate plot.vsel() to identify if the search was ",
+              "terminated too early. If this is the case, run variable ",
+              "selection with larger value for `nterms_max`.")
```
avehtari: maybe
->
              "Investigate plot.vsel() to identify if the search was ",
              "terminated too early or the criterion is too strict. If this is the case, run variable ",
              "selection with larger value for `nterms_max` or use less strict criterion.")

avehtari (approved):

PR #556, opened 2026-01-19
use relative fontsize for model size numbers

### LINKED ISSUE #555: ggplot2 4.x changes the plot.vsel model size font size
I use 
```
theme_set(bayesplot::theme_default(base_family = "sans", base_size = 16))
```

With ggplot2 3.5.x plot.vsel looked like

Image

With ggplot2 4.x plot.vsel looks like

Image

Everything else is the same except for the model size labels are too big. This is related to ggplot2 4.x change so that the default font size for `geom_text()` (and other geoms) is now inherited from `theme(geom = element_geom())`, while previous default was hard coded to size 11. Instead of fixing the font size to the previous fixed size 11, I think it is better to make it depend on `base_size` like all the other text fonts in the figure. In my plot 11/16 = 0.69*`base_size` looks quite nice and it seems 0.7 is common scaling factor and thus I suggest using 0.7*`base_size`.

avehtari:
Fixes #555 

This would add requirement that ggplot2 version needs to be at least 4.*

### DIFF
--- a/R/methods.R
+++ b/R/methods.R
@@ -1105,9 +1105,15 @@ plot.vsel <- function(
       x_color_txt <- "black"
     }
     pp <- pp +
-      geom_text(aes(y = -Inf, label = .data[["size_chr"]],
-                    size = from_theme(fontsize * 0.7)),
-                vjust = -0.5, color = x_color_txt)
+      if (packageVersion("ggplot2") < "4.0.0") {
+        geom_text(aes(y = -Inf, label = .data[["size_chr"]]), vjust = -0.5,
+                        color = x_color_txt)
+      } else {
+        geom_text(aes(y = -Inf, label = .data[["size_chr"]],
+                      size = from_theme(fontsize * 0.7)),
+                  vjust = -0.5, color = x_color_txt)
+
+      }
   }
   # Miscellaneous stuff (axes, theming, faceting, etc.):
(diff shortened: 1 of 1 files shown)


avehtari:
@jgabry I'm asking your opinion on this PR. See also the issue for more explanation

R/methods.R
```diff
@@ -1105,8 +1105,9 @@ plot.vsel <- function(
       x_color_txt <- "black"
     }
     pp <- pp +
-      geom_text(aes(y = -Inf, label = .data[["size_chr"]]), vjust = -0.5,
-                color = x_color_txt)
+      geom_text(aes(y = -Inf, label = .data[["size_chr"]],
+                    size = from_theme(fontsize * 0.7)),
+                vjust = -0.5, color = x_color_txt)
```
aloctavodia: Do we want this instead?

```suggestion
      if (packageVersion("ggplot2") < "4.0.0") {
        geom_text(aes(y = -Inf, label = .data[["size_chr"]]), vjust = -0.5,
                        color = x_color_txt)
      } else {
        geom_text(aes(y = -Inf, label = .data[["size_chr"]],
                      size = from_theme(fontsize * 0.7)),
                  vjust = -0.5, color = x_color_txt)

      }
```

avehtari: I'm fine with adding the version check

aloctavodia (approved):

